PR #22, opened 2017-10-24
Feature/compile for npm

domharrington:
Can you review this? May not be ready to merge yet but seems to be working locally.


mjcuva:
Seems good to me! It would be really nice to have something that could just run all of the commands on whatever is on npm as a last layer of testing

PR #23, opened 2017-11-08
Support sending files to build services

mjcuva:
Added support for sending files to build services.

- Posts as multipart/form-data
- Added api.file, which takes a path to a file
- Parses response to handle files being returned, as well as object responses

## Todo

- [x] Add support for urls in api.file
- [x] `api local`
- [x] `api link`


api.js
```diff
@@ -4,76 +4,88 @@ require('colors');
...
 
-    const event = {
-      name: action,
-      data,
-      pathOverride: localLinks.services[this.service],
-      errors,
-    };
+    // If no data is passed in, default to {}
+    if (typeof data === 'function') {
+      callback = data;
```
domharrington: It looks as though we've lost some coverage here... did this used to be covered?

commands/local.js
```diff
@@ -40,6 +30,8 @@ module.exports.run = (args) => {
     if (err) {
       const parsedError = JSON.parse(err);
       console.log(parsedError);
+    } else if (Buffer.isBuffer(response)) {
+      process.stdout.write(response);
```
domharrington: Any way to test this? Could monkey path over `process.stdout.write` for the test? Is this any different that doing `console.log(response)`?

commands/run.js
```diff
@@ -56,7 +44,11 @@ module.exports.run = (args, opts) => {
     }
 
     return invoke(team.key, service, action, data, true).then((response) => {
-      console.log(response.body);
+      if (Buffer.isBuffer(response.file)) {
+        process.stdout.write(response.file);
```
domharrington: Same as above. Maybe this piece can be abstracted out, might make it easier to test.

test/utils/file-utils.test.js
```diff
@@ -0,0 +1,110 @@
...
+const fileUtils = require('../../utils/file-utils');
+
+describe('file-utils', () => {
+  describe('#isUrl', () => {
+    it('should return true if valid url', () => {
+      assert(fileUtils.isUrl('https://google.com'));
+    });
+
+    it('should return false if not valid url', () => {
+      assert.equal(false, fileUtils.isUrl('/path/to/image'));
```
domharrington: This should be the other way around `assert.equal(fileUtils.isUrl('/path/to/image'), false);`

utils/file-utils.js
```diff
@@ -0,0 +1,70 @@
+const fileType = require('file-type');
+const Stream = require('stream');
+const request = require('request-promise');
+const fs = require('fs');
+
+exports.file = (p) => {
+  if (exports.isUrl(p)) {
+    return request.get({ url: p, encoding: null });
```
domharrington: Anyway to write a test for this and mock out the response?

utils/file-utils.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+  return fs.createReadStream(p);
+};
+
+exports.isUrl = (s) => {
+  const regexp = /(ftp|http|https):\/\/(\w+:{0,1}\w*@)?(\S+)(:[0-9]+)?(\/|\/([\w#!:.?+=&%@!\-/]))?/;
```
domharrington: Can you use http://nodejs.org/dist/latest/docs/api/url.html#url_class_url instead for this? Attempt to construct a `new URL` and add a try..catch to see if it succeeds?

utils/file-utils.js
```diff
@@ -0,0 +1,70 @@
...
+  return regexp.test(s);
+};
+
+// Converts a stream to a buffer
+// Useful for locally calling a service
+// with a file
+exports.streamToBuffer = (s) => {
+  return new Promise((resolve) => {
+    const out = [];
+    s.on('data', (data) => {
```
domharrington: I think we need to add an error case to this:

```js
s.on('error', reject)
```

utils/file-utils.js
```diff
@@ -0,0 +1,70 @@
...
+      out.push(data);
+    });
+    s.on('end', () => {
+      resolve(Buffer.concat(out));
+    });
+  });
+};
+
+// Converts buffers in response to file type
+exports.parseFileResponse = (response) => {
```
domharrington: Not sure what to do to simplify this. Maybe pull that parse reviver function outside of this function into a named one? Wondering if you came across this to do the buffer part? https://github.com/jprichardson/buffer-json Ours looks like it does more than that though....

utils/handler-utils.js
```diff
@@ -76,3 +76,15 @@ module.exports.log = function log() {
 };
 
 module.exports.getSecret = secrets => secret => secrets.find(s => s.key === secret).value;
+
+module.exports.fixBuffers = (d) => {
```
domharrington: What's this for?

utils/utils.js
```diff
@@ -89,9 +92,35 @@ Don't have an account? Signup is free and takes 5 seconds!
...
 };
 
+// Converts command line arguments to a javascript object
+// test=1 test2='hello' test3=@/path/to/file
+// { test: 1, test2: 'hello', test3: BUFFER }
+exports.parseArgs = (args) => {
+  const data = {};
+  const passedData = exports.fixMinimistSpaces(args);
+  for (const arg of passedData) {
+    const i = arg.indexOf('=');
```
domharrington: Can you pull this whole chunk into a separate part to simplify the complexity?

utils/utils.js
```diff
@@ -155,3 +184,31 @@ exports.getUnchangedDocs = (docs) => {
   return docs.filter(doc => doc.fullDescription && doc.fullDescription.indexOf('https://docs.readme.build/docs/writing-documentation') >= 0)
   .map(doc => doc.name);
 };
+
+// Parses response to make sure its the correct type
+exports.parseResponse = (response) => {
+  let parsedResponse = response.body;
+  try {
+    parsedResponse = JSON.parse(response.body, (k, v) => {
+      return v && v.type === 'Buffer' ? Buffer.from(v.data) : v;
```
domharrington: Is this the same as https://github.com/readmeio/api/pull/23/files#diff-d6c8902e1266636417b54018791967d1R34?

commands/local.js
mjcuva: `console.log` formats the output, so for writing binary data it doesn't work. `process.stdout.write` just dumps the output as is

utils/handler-utils.js
mjcuva: When a file gets passed to lambda, it gets stringified into that weird buffer string format, so we have to convert it back to being a buffer

utils/utils.js
mjcuva: They do slightly different things, but I think I should combine them

api.js
mjcuva: I'm pretty sure it never was

PR #24, opened 2018-01-10
Add output functionality

iMuzz:


src/lib/invoke.js
```diff
@@ -9,6 +11,12 @@ module.exports = (key, service, action, data, isCLI) => {
     'X-Build-Meta-OS': `${os.type().toLowerCase()}@${os.release()}`,
   };
 
+  if (data['x-build-outputs'] !== undefined) {
```
mjcuva: Instead of passing it through `data['x-build-outputs']`, why not just pass it as a param into invoke?

src/api.js
```diff
@@ -9,7 +9,7 @@ const invoke = require('./lib/invoke');
 
 const linksPath = path.join(utils.sharedDirectoryPath(), 'links.json');
 
-module.exports.run = (action, d, cb) => {
+module.exports.run = (action, d, outputs, cb) => {
```
domharrington: Doesnt this mean that if you want to use `run` with a `cb` that you *have* to pass in something for outputs?

```js
const api = require('api').config('domh_ccd5142d2107c48ce32f75');

api('build-test')
  .run('helloWorld', { name: 'Name' }, 
   // this has to be here now? 
   {},
   function cb () {})
```

Can you make it optional?

src/lib/invoke.js
```diff
@@ -9,6 +11,12 @@ module.exports = (key, service, action, data, isCLI) => {
...
+    const parsedOutput = querystring.stringify(data['x-build-outputs']);
+    delete data['x-build-outputs'];
+    headers['X-Build-Output'] = parsedOutput;
```
domharrington: Can you add a test here to make sure the header gets set? https://github.com/readmeio/api/blob/master/test/commands/run.test.js

src/lib/invoke.js
iMuzz: Yeah I thought about that.. 

But the [last parameter](https://github.com/readmeio/api/blob/master/src/lib/invoke.js#L5) is [optional.](https://github.com/readmeio/api/blob/master/src/api.js#L78).

I could make it so `invoke` takes another param. But then I'd have to change [this line](https://github.com/readmeio/api/blob/master/src/api.js#L78) to be: 
```
return invoke(scope.key, scope.service, action, data, undefined, outputs).then((response) => {
```

Long story short: 
1. I didn't want to add a parameter and then have to pass in `undefined` there. 
2. I didn't want to change the order of the parameters coming in.

But if we really don't like this.. I'm happy to change it so that we pass in this object so that there isn't an order dependency anymore!
```
{ 
  key: '',
  service: '',
  action: '' ,
  data: '',
  isCLI : ''
}
```

mjcuva: What about doing this: 

```node
invoke(key, service, action, data, opts)
```

And making ops have `isCLI`, `outputs` and whatever else we decide to add in the future? 

I'm also not against just making it an object.

src/api.js
iMuzz: @domharrington fixed!

src/lib/invoke.js
iMuzz: @domharrington is this file for the `.run` function?

src/lib/invoke.js
domharrington: I quite like just making it an object.

PR #28, opened 2020-05-08
SDK Generation

erunion:
## 🧰 What's being changed?

This completely rewrites the existing `api` module to focus on the autogeneration of SDKs from an OpenAPI 3.x document. For example:

```js
const sdk = require('./src')('https://raw.githubusercontent.com/readmeio/oas/master/packages/examples/3.0/yaml/readme.yaml');

sdk
  .auth('readmeApiToken')
  .getChangelogs({
    perPage: 10,
    page: 1
  })
  .then(res => res.json())
  .then(res => {
    console.log(`there are ${res.length} changelogs`);
    console.log(res[0]);
  });
```

## 📓 How does it work?
[Proxies!](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) Proxies are a recent addition to JS that allows for one to create wildcard-y object accessors. With this we can create an SDK proxy that has the base methods all SDKs have (like `.auth()`), and then automatically allow for the end-user to access operations like `.getChangelogs()` or `.get()` on the chain.

What'll happen when a user first makes a request via `.getChangelogs()` is that we'll fetch the OpenAPI definition they supplied, process it, cache it into `node_modules/.cache/api`, and run through the definition to add any present operation IDs as chainable methods (like `.getChangelogs()` on the Proxy. On subsequent requests, the OpenAPI definition is re-fetched from the local cache, and re-added to the Proxy.

It's like magic.

## 🎙How do I make requests?
With the chained operation IDs, or HTTP verb methods like GET or POST, that are automatically generated onto the SDK, you can pass them parameters and request body payloads by supplying either a `body` object or array, and/or a `metadata` (parameters) object. The SDK will then determine, according to the OpenAPI definition, what goes where. For example:

1. Just `metadata`: `sdk.showPetById({ petId: 1234 }).then(...)`
2. Just `body`: `sdk.createPets({ name: 'Buster' }).then(...)`
3. Both: `sdk.updatePet({ name: 'Buster 2' }, { petId: 1234 }).then(...)`

## 🧪 Testing

To test out the SDK, give it an OpenAPI URL or local file and go wild.


.gitignore
```diff
@@ -1,2 +1,3 @@
 coverage/
 node_modules/
+example.js
```
domharrington, line 3: I think it's fine to include an example.js file in a repo! Could just be to petstore, or w/e

package.json
```diff
@@ -16,11 +17,31 @@
...
   "author": "Jon Ursenbach <jon@readme.io>",
   "license": "ISC",
-  "dependencies": {},
+  "engines": {
+    "node": ">=10"
+  },
+  "dependencies": {
+    "@apidevtools/json-schema-ref-parser": "^9.0.1",
+    "@openapi-contrib/openapi-schema-to-json-schema": "^3.0.0",
+    "@readme/oas-examples": "^3.4.0",
```
domharrington, line 26: Lets shift over the correct ones to devDependencies that dont need to be real deps.

src/lib/prepareParams.js
```diff
@@ -0,0 +1,134 @@
...
+  // body and pass it along.
+  if (!('body' in params) && typeof metadata === 'undefined') {
+    if (!hasDigestedParams) {
+      // No parameters were able to be digested, so we just have to assume that what the user supplied was for a body.
+      // This might lead to unwanted false positives if an OAS isn't accurate, but short of throwing an error there
+      // isn't anything we can really do about it.
+      params.body = body;
+    } else {
+      const intersection = Object.keys(body).filter(value => Object.keys(digested).includes(value)).length;
+      if (intersection && (intersection / Object.keys(body).length) > 0.25) {
```
domharrington, line 72: Would love to see some tests around this logic.

erunion:
@gkoberger @domharrington I decided to sleep on it and I'm going to stick with the arg format as `body, metadata`. 

A major portion of API requests will only constitute `metadata`, and making a request with `get(metadata)` will work as expected. And for those that require both, the `body` that's supplied is the more important part of the request and makes sense to be the primary argument and take precedence over `metadata`.

src/lib/prepareParams.js
erunion: Done!

README.md
```diff
@@ -1,2 +1,125 @@
...
+With the exception of OpenID, it supports all forms of authentication supported by the OpenAPI specification! Just give `.auth()` your credentials and it'll figure out how to use it according to the API you're using.
+
+For example:
+
+* HTTP Basic auth: `sdk.auth('username', 'password')`
+* Bearer tokens (HTTP or OAuth 2): `sdk.auth('myBearerToken')`
+* API Keys: `sdk.auth('myApiKey')`
+
+### Parameters and Payloads
+When supplying parameters and/or request body payloads to an API request, you don't to explicitly define what goes where since the API definition contains all that information. All you need to do is supply either one or two objects:
```
kanadgupta, line 50: I _think_ this is the omitted word here?

```suggestion
When supplying parameters and/or request body payloads to an API request, you don't need to explicitly define what goes where since the API definition contains all that information. All you need to do is supply either one or two objects:
```

README.md
```diff
@@ -1,2 +1,125 @@
...
+
+* `body`: This will contain all data required for a request body payload for a POST, PUT, etc. request. It can either be an array or an object, whichever you need to use the API operation you're using.
```
kanadgupta, line 52:
```suggestion
* `body`: This will contain all data required for a request body payload for a POST, PUT, etc. request. It can either be an array or an object — whichever you need to use the API operation you're using.
```

README.md
```diff
@@ -1,2 +1,125 @@
...
+
+Since `petId` matches up with the `petId` path parameter, the SDK here will issue a GET request against `/pets/1234`.
+
+What about a POST request?
+
+```js
+sdk.createPets({ name: 'Buster' }).then(...)
+```
+
+Since `name` here would coorespond on `createPets` to request body payload, this will issue a POST request against `/pets` to make a new pet named "Buster".
```
kanadgupta, line 69:
```suggestion
Since `name` here would correspond on `createPets` to request body payload, this will issue a POST request against `/pets` to make a new pet named "Buster".
```

.github/workflows/ci.yml
```diff
@@ -19,8 +19,8 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
 
-    - name: npm install
+    - name: Install dependencies
       run: npm ci
 
-    - name: tests
+    - name: Run tests
```
domharrington, line 25: Can we get code coverage sent up to codecov.io?

README.md
```diff
@@ -1,2 +1,125 @@
-# api
+# 🚀 api
```
kanadgupta, line 1: whoooooosh

__tests__/cache.test.js
```diff
@@ -0,0 +1,198 @@
...
+const Cache = require('../src/cache');
+const pkg = require('../package.json');
+
+let readmeExampleJson;
+let readmeExampleYaml;
+const examplesDir = join(__dirname, 'examples');
+
+expect.extend({
+  // Custom matcher so we can easily test that dereferencing of OpenAPI files is working as expected.
+  toBeDereferenced(received) {
```
domharrington, line 17: This is really nifty!

__tests__/cache.test.js
```diff
@@ -0,0 +1,198 @@
...
+    return {
+      message: () => `expected supplied array to be dereferenced`,
+      pass: false,
+    };
+  },
+});
+
+beforeEach(async () => {
+  readmeExampleJson = await fs.readFile(
+    join(__dirname, '../node_modules/@readme/oas-examples/3.0/json/readme.json'),
```
domharrington, line 35: You should be able to do this using a `require.resolve('@readme/oas-examples/3.0/json/readme.json')`

__tests__/index.test.js
```diff
@@ -0,0 +1,290 @@
...
+      'head',
+      'patch',
+      'trace',
+      'list-data-sets',
+      'list-searchable-fields',
+      'perform-search',
+    ]);
+  });
+
+  it.todo('should error if passing in swagger 2');
```
domharrington, line 89: Sounds like all of these can be removed now they've been implemented in the cache?

erunion:
@domharrington Proxy is supported currently in Node 6+, every non-IE browser, and Safari (though it doesn't support `getPrototypeOf`, but I'm not using that here). The SDK could likely be reconfigured to work in a browser environment, but that's going to require some thought into how it handles file caching since it relies on the `fs` module.

__tests__/index.test.js
erunion: Yeah this and a couple other todos can get removed.

README.md
erunion: We need an astronaut Owlbert.

__tests__/cache.test.js
erunion: TIL about `require.resolve`

PR #49, opened 2020-05-25
Beautified this JavaScript file.

Katheesh:
Not changed program. because I changed syntax showing properly. thanks

## 🧰 What's being changed?

Describe in detail what this PR is for.

## 🧪 Testing

* [ ] **🆕 I'm adding something new!**
* [ ] **🐛 I'm fixing a bug!**


erunion:
Hi! Thanks for the submission, but fortunately we're using Prettier and the file is already formatted according to its ruleset.

PR #50, opened 2020-05-26
fix(package.main): specify the entry point file

rafegoldberg:
## 🧰 What's being changed?

When I install and require the API package from my code, I get the following error:

```
internal/modules/cjs/loader.js:628
    throw err;
    ^
Error: Cannot find module 'api'
```

I can get it working by simply setting the required path to `api/src/index.js`. This PR adds a main field, which will hopefully make this workaround moot.

## 🧪 Testing

Install the API package in a local directory and try requiring the dependency in a node script. You should see the aforementioned "module not found" error.

## 🗳 Checklist

* I'm fixing a bug

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "api",
   "version": "2.0.0",
   "description": "Generate an SDK from an OpenAPI definition",
+  "main": "src/index.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


erunion:
pubbed to 2.0.1

Issue #131, opened 2020-08-06
Support multipart/form-data operations


erunion:
![Screen Shot 2020-08-06 at 4 18 30 PM](https://user-images.githubusercontent.com/33762/89592028-83241100-d800-11ea-8af0-ea52c72d5446.png)

I made a bunch of changes in https://github.com/readmeio/httpsnippet/pull/1 to allow for Node `request` snippets with params that have `fileName` to surface that instead of the data URL, we should copy those changes over here into our code snippet for the SDK.

![Screen Shot 2020-08-06 at 4 19 42 PM](https://user-images.githubusercontent.com/33762/89592103-a77fed80-d800-11ea-9b25-0bfb726dfee8.png)

erunion:
Snippets should look like:

```
sdk.post('/anything', {
  documentFile: 'owlbert.png',
})
  .then(res => res.json())
  .then(res => {
    console.log(res);
  });
```

The SDK should handle creating read streams for `fetch` instead of offloading that work to the user, but for the sake of magic we should support passing streams too.

erunion:
[fetch-har@4.0.0](http://npm.im/fetch-har) has support for multipart/form-data so this should pull that in.

Issue #130, opened 2020-08-04
Support multipart/form-data operations


erunion:
Operations with `multipart/form-data` content types are currently being generated with snippets that look like the following:

```js
const sdk = require('api')('https://example.com/openapi.json');
    
sdk.post('/multipart', {orderId: '10'}, {'content-type': 'multipart/form-data; boundary=---011000010111000001101001'})
  .then(res => res.json())
  .then(res => {
    console.log(res);
  });
```

Since the SDK is already aware that the operation is `multipart/form-data`, there should be no need to add the `Content-Type` header into the snippet.

erunion:
Closing this in favor of https://github.com/readmeio/api/issues/131 as it's a better ticket.

PR #132, opened 2020-08-08
feat: support multipart/form-data

erunion:
## 🧰 What's being changed?

* [x] Adds general support for `multipart/form-data` requests and the ability to accept file uploads.
* [x] Updates SDK snippets to comply with these changes.

In order to upload a file as part of a `multipart/form-data` request (or one of its supported variants), you can just give a path to the file as the value of the property in the parameters for the request. For example:

```js
sdk.post('/anything', { documentFile: 'owlbert.png' })
  .then(res => res.json())
  .then(res => {
    console.log(res);
  });
```

This will automatically locate `owlbert.png`, read it into memory, convert it into [data URL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs), and hand it off to [fetch-har](https://www.npmjs.com/package/fetch-har).

You can also give it a stream, and the SDK will handle everything as well like magic.

```js
sdk.post('/anything', { documentFile: fs.createReadStream('owlbert.png') })
  .then(res => res.json())
  .then(res => {
    console.log(res);
  });
```

Resolves https://github.com/readmeio/api/issues/131 and https://github.com/readmeio/api/issues/130


packages/api/src/lib/prepareParams.js
```diff
@@ -77,6 +84,58 @@ module.exports = function (operation, body, metadata) {
...
+  // @todo support content types like `image/png` where the request body is the binary
+
+  // If the operation is `multipart/form-data`, or one of our recognized variants, we need to look at the incoming
+  // body payload to see if anything in there is either a file path or a file stream so we can translate those into a
+  // data URL for `@readme/oas-to-har` to make a request.
+  if ('body' in params && operation.isMultipart()) {
+    const schema = getSchema(operation, operation.oas) || { schema: {} };
+    const bodyKeys = Object.keys(params.body);
+
+    // Loop thorugh the schema to look for `binary` properties so we know what we need to convert.
```
erunion, line 96: Through*

packages/api/src/lib/prepareParams.js
```diff
@@ -77,6 +84,58 @@ module.exports = function (operation, body, metadata) {
...
+    const conversions = [];
+    Object.keys(schema.schema.properties)
+      .filter(key => schema.schema.properties[key].format === 'binary')
+      .filter(x => bodyKeys.includes(x))
+      .forEach(async prop => {
+        let file = params.body[prop];
+        if (typeof file === 'string') {
+          file = path.resolve(file);
```
erunion, line 104: Add comment explaining why were doing this.

erunion:
Holding off on merging this in until next week because after running some tests against our own API, `fetch-har` still needs some work.

```
{ error: 'SPEC_INVALID',
  message:
   'The spec you uploaded isn\'t a valid JSON or YAML file. Here\'s the error we found: Not valid YAML',
  suggestion:
   'Try running your spec through a validator to see if they pick up any issues with the formatting!',
  docs:
   'https://docs.readme.com/developers/logs/********',
  help:
   'If you need help, email support@readme.io and include the following link to your API log: \'https://docs.readme.com/developers/logs/********\'.',
  poem:
   [ 'When we asked for your swagger',
     'It seems you misunderstood',
     'You sent us a cool attitude',
     'Instead some JSON would have been good' ] }
```

Turns out sending a data URL instead of file contents doesn't actually work in real-world scenarios.

PR #154, opened 2020-09-08
fix: if no cache dir is determined, fallback to the os temp dir

### LINKED ISSUE #107: TypeError: cacheDir is not a function
I got this error when testing out the API module on our reference guide: https://docs.readme.com/developers/reference/projects

```js
const sdk = require('api')('https://docs.readme.com/developers/openapi/5e1a4f6284248b0018149174');

sdk.getProject()
  .then(res => res.json())
  .then(res => {
    console.log(res);
  });
```

I saved it to /tmp/node-simple.js and ran `npm install api`, then ran `node node-simple.js` and got the following error:

```sh
$ node node-simple.js
/private/tmp/node_modules/api/src/cache.js:17
    this.dir = cacheDir();
               ^

TypeError: cacheDir is not a function
    at new SdkCache (/private/tmp/node_modules/api/src/cache.js:17:16)
    at Sdk.load (/private/tmp/node_modules/api/src/index.js:31:19)
    at module.exports.uri (/private/tmp/node_modules/api/src/index.js:141:23)
    at Object.<anonymous> (/private/tmp/node-simple.js:1:27)
    at Module._compile (internal/modules/cjs/loader.js:776:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:787:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:829:12)
```

Confusingly when it's in another directory it works as intended? Could this be an error in running from /tmp? I use it exclusively for messing around with stuff like this.

erunion:
## 🧰 What's being changed?

This resolves #107 and edge case with the [find-cache-dir](https://www.npmjs.com/package/find-cache-dir) module where if it can't determine a cache directory, it'll return `undefined`. When this could crop up is if a user runs `npm install api` in a fresh directory without a `package.json` file. Since `find-cache-dir` looks for the root-most directory with a `package.json` file to determine where to put its `.cache/` directory into `node_modules/`, it can't.

So to account for this edge case, if we end up with an undefined cache directory from the module we'll instead fallback to creating a temp directory at the OS level.

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -33,6 +33,7 @@
     "form-data": "^3.0.0",
     "get-stream": "^6.0.0",
     "js-yaml": "^3.14.0",
+    "make-dir": "^3.1.0",
     "mimer": "^1.1.0",
     "node-fetch": "^2.6.0"
   },
--- a/packages/api/__tests__/cache-tmp.test.js
+++ b/packages/api/__tests__/cache-tmp.test.js
@@ -1,0 +1,17 @@
+jest.mock('find-cache-dir', () => {
+  return () => undefined;
+});
+
+const os = require('os');
+const Cache = require('../src/cache');
+
+// Since this test is mocking out the `find-cache-dir` module for a single test, it needs to be run separately from the
+// rest of the cache tests, otherwise all of those tests would use this mocked out version.
+test('should fallback to an os-level temp directory if a cache directory cannot be determined', () => {
+  const dir = os.tmpdir();
+  const cacheStore = new Cache('http://example.com/readme.json');
+
+  expect(cacheStore.dir).toMatch(dir);
+  expect(cacheStore.cacheStore).toMatch(dir);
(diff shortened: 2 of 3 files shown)


packages/api/package.json
```diff
@@ -33,6 +33,7 @@
     "form-data": "^3.0.0",
     "get-stream": "^6.0.0",
     "js-yaml": "^3.14.0",
+    "make-dir": "^3.1.0",
```
erunion, line 36: This module is already coming in via `find-cache-dir`, so I'm using that instead of loading in something else like [mktemp](https://www.npmjs.com/package/mktemp).

domharrington (approved):

PR #240, opened 2021-02-03
feat: automatically parse the api response based on content-type 

domharrington:
## 🧰 What's being changed?

This allows us to remove the `.then(res => res.json())` line from our
code samples.

```js
const sdk = require('api')('@readme/5f7ceacf43621b0311080a58');

sdk.getAPISpecification({perPage: '10', page: '1'})
```

We only perform the parsing if it's a successful response. If there's an error
case then we return with the original response object so you can access
the status as well as the actual body.

The code and tests for this have been stolen verbatim from the explorer:
https://github.com/readmeio/api-explorer/blob/77b90ebed4673f168354cdcd730e34b7ee016360/packages/api-explorer/src/lib/parse-response.js#L13-L30

BREAKING CHANGE: this is a breaking change.

### DIFF
--- a/packages/httpsnippet-client-api/src/index.js
+++ b/packages/httpsnippet-client-api/src/index.js
@@ -245,7 +245,7 @@ module.exports = function (source, options) {
 
   code
     .push(`sdk${accessor}(${args.join(', ')})`)
-    .push(1, '.then(json => console.log(json))')
+    .push(1, '.then(res => console.log(res))')
     .push(1, '.catch(err => console.error(err));');
 
   return code.join();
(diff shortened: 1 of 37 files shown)


packages/httpsnippet-client-api/src/index.js
```diff
@@ -245,7 +245,6 @@ module.exports = function (source, options) {
 
   code
     .push(`sdk${accessor}(${args.join(', ')})`)
-    .push(1, '.then(res => res.json())')
     .push(1, '.then(json => console.log(json))')
```
erunion, line 248: For endpoints that return non-JSON content, what should we do here? Change this part of the snippet to `.push(1, '.then(res => console.log(res))')` and only mention "json" if the SDK is doing this autoretrieval of `res.json()`?

domharrington: Fix we spoke about on Slack https://github.com/readmeio/api/pull/240/commits/4e654ae541269266b6b6a81d34bd9b0e10351bcb

erunion (approved):

PR #253, opened 2021-03-16
chore: update deps and require npm@7

erunion:
## 🧰 What's being changed?

* [x] Dropped support for Node 10. It's currently in maintenance mode and is being EOL'd in April.
* [x] Upgrades the local npm usage to require npm@7 for all local development.
* [x] Updates `@readme/httpsnippet` to the latest release and moves it from being a dep to a peerDep.
* [x] Updates `@readme/oas-to-har` to the latest release to pull in some fixes w/r/t `format: json` request bodies. https://github.com/readmeio/api-explorer/pull/1197 
* [x] Updates `oas` to the latest version across both `api` and `httpsnippet-client-api`.
  * I added a new `getAuth` method into `Oas` (https://github.com/readmeio/oas/pull/376) but we've been unable to take advantage of it because we have some semver clases in ReadMe and it's preferring 10.0.0 over 10.2.0 because `httpsnippet-client-api` had 10.0.0 listed. Incredibly frustrating, but this should let us fully upgrade `oas` everywhere.
* [x] Moves `oas` to being a peerDep in `httpsnippet-client-api`. Since `httpsnippet-client-api` is a bit more volatile than `api`, it makes sense to move it to being a peerDep so we can have a bit more leeway when installing it in other other packages and frees us up from having to always do version bumps on `oas`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,10 +2,8 @@
   "name": "api-monorepo",
   "private": true,
   "scripts": {
-    "bootstrap": "lerna bootstrap --hoist",
     "clean": "lerna clean",
     "lint": "lerna run lint --stream",
-    "postinstall": "npm run bootstrap",
     "publish": "lerna publish",
     "test": "lerna run test --stream",
     "version": "conventional-changelog --pkg lerna.json -i CHANGELOG.md -s && git add CHANGELOG.md"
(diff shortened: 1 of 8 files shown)


packages/api/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "src/index.js",
   "scripts": {
     "lint": "eslint .",
-    "pretest": "npm run lint && npm run prettier",
```
erunion, line 8: I've started removing these wherever I see them because our ESLint config already runs Prettier, so this is always going to run Prettier on an already prettified codebase.

package.json
```diff
@@ -15,8 +13,12 @@
     "url": "https://github.com/readmeio/api.git"
   },
   "engines": {
-    "node": ">=10"
+    "node": "^12 || ^14",
+    "npm": "^7"
```
erunion, line 17: @todo remove this because it might prevent people on npm@6 from installing the packages

domharrington (approved):
LGTM!

PR #264, opened 2021-03-25
feat: add a config() function that allows disabling of response parsing

domharrington:
## 🧰 What's being changed?

If you ever need to access the raw Response object (instead of us parsing
automatically) you can do the following:

```js
sdk.config({ parseResponse: false });
sdk.deletePet({ id: petId });
```

This follows a similar pattern to the auth function.

### DIFF
--- a/packages/api/README.md
+++ b/packages/api/README.md
@@ -24,7 +24,7 @@ Using `api` is as simple as supplying it an OpenAPI and using the SDK as you wou
 ```js
 const sdk = require('api')('https://raw.githubusercontent.com/readmeio/oas/master/packages/examples/3.0/json/petstore.json');
 
-sdk.listPets().then(res => res.json()).then(res => {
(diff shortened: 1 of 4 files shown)


packages/api/README.md
```diff
@@ -24,7 +24,7 @@ Using `api` is as simple as supplying it an OpenAPI and using the SDK as you wou
 ```js
 const sdk = require('api')('https://raw.githubusercontent.com/readmeio/oas/master/packages/examples/3.0/json/petstore.json');
 
-sdk.listPets().then(res => res.json()).then(res => {
+sdk.listPets().then(res => {
```
erunion, line 27: Starting to wonder if we should update `res` in our code snippets to be `json` so it's clearer that you're getting JSON back and not a `Response` object.

erunion (approved):

PR #284, opened 2021-05-11
feat: adding support for server variables

erunion:
## 🧰 What's being changed?

This adds handling for alternate servers and server variables into a new `.server()` method.

* [x] Adds support for handling server variables to `api`.
* [x] Adds support for adding server variables to generated snippets.
* [x] Fixes a bug in our handling of paths with path parameters where the path parameters might not always be in the snippet.
* [x] Cleaning up some tests that used a `createOas` fixture in favor of using fuller OpenAPI examples from [@readme/oas-examples](https://npm.im/@readme/oas-examples).

## 🐳  How does it work?

You can either pass in a full server URL into `server`:

```js
sdk.server('https://example.com/var1/var2')
```

Or if you instead want to pass in individual server variable params, you can do so as well:

```js
sdk.server({'https://example.com/{pathParam1}/{pathParam2}', {
  pathParam1: 'var1',
  pathParam2: 'var2',
})
```

By default, snippets generated with `httpsnippet-client-api` will be generated with the former syntax.


gkoberger:
Hmmm okay I like this enough... it's clunky, but there's not much we can really do.

I'd love to eventually figure out a way to add shorthand, but I can't think of any way to do that in a deterministic way. (Especially for non-variable servers where, for example, one is a test server and one is real.)

It would be nice if eventually (no need to do this now) we could let people specify a x-readme-default server and if there's no variables then we default to that and don't make people spell it out?

erunion:
> It would be nice if eventually (no need to do this now) we could let people specify a x-readme-default server and if there's no variables then we default to that and don't make people spell it out?

We kind of handle this already by treating the first server in `servers` as their default when we run `oas.url()`: https://github.com/readmeio/oas/blob/main/tooling/index.js#L148-L151

> I'd love to eventually figure out a way to add shorthand, but I can't think of any way to do that in a deterministic way. (Especially for non-variable servers where, for example, one is a test server and one is real.)

We could potentially do this as `sdk.config({server: 'dev'})` and base it off the `description` in the [Server Variable Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#server-variable-object) but the problem with that field is that its fulltext Markdown so for one user it might be `dev`, another it might be: 

```js
sdk.config({ server: 'Our local development server that runs on localhost:3000' });
```

We could also treat the servers by their index in the `servers` array, but that makes the request really opaque:

```js
/* oas.servers: [
  { url: 'prod' },
  { url: 'dev' }
] */

sdk.config({ server: 1 });
``

**Comment by gkoberger** [comment]
Oh great!

**Comment by erunion** [comment]
> I don't like the last one (if no other reason, it's possible a swagger creator could switch the order w/o thinking about it, and break tons of APIs!)

Agreed

**Comment by erunion** [comment]
@gkoberger What about changing this to:

```js
sdk.server('http://api.example.local:3000/api/')
```

and:

```js
sdk.server('https://example.com/{pathParam1}/{pathParam2}', {
  pathParam1: 'var1',
  pathParam2: 'var2',
})
```

Dashron:
Personally I'm always a fan of explicit config methods such as `server`. It's nicer in IDEs, can lead to clearer docs, and clearer arguments. It can lead to bloat, but `server` is important enough that it's probably worth it.

packages/api/__tests__/config.test.js
```diff
@@ -2,29 +2,89 @@ const nock = require('nock');
...
+        },
+      });
+
+      return sdk.post('/').then(res => {
+        expect(res).toStrictEqual(response);
+        mock.done();
+      });
+    });
+
+    it.todo('should be able to supply a url on an OAS that has no servers defined');
```
Dashron, line 72: Did you mean to keep these as todos?

Dashron (approved):
just one small comment

packages/api/__tests__/config.test.js
erunion: Yeah, these cases are going to require some overhauls to `@readme/oas-to-har` to allow for a custom base URL to be passed in that can't be surfaced with `oas.url()`.

Dashron: 👍

PR #380, opened 2022-01-27
feat: improved handling of file uploads

erunion:
| 🚥  | Fix RM-3239, RM-3241, RM-3240 |
| :-- | :--- |

## 🧰 What's being changed?

* [x] 🆕 Upgrades [fetch-har](https://npm.im/fetch-har) to incorporate our new rewritten handling of the [FormData API](https://developer.mozilla.org/en-US/docs/Web/API/FormData).
  * ICYMI check out [fetch-har#237](https://github.com/readmeio/fetch-har/pull/237) for the bulk of where this rewrite happened.
* [x] 🆕 Swapping out [node-fetch](https://npm.im/node-fetch) for [isomorphic-fetch](https://npm.im/isomorphic-fetch) for no real reason other than `isomorphic-fetch` handles updating the [globalThis](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) state for us.
* [x] 🆕 Adds support for top-level binary payload request bodies like `image/png` or any request body schema that is just `type: string; format binary`.
* [x] 🧹 Changes how we access request bodies out of OAS files to using a new `getRequestBody` accessor in [oas](https://npm.im/oas).
  * Also changes how we trawl through request bodies to reference our generated JSON Schema (using `getParametersAsJsonSchema()` in [oas](https://npm.im/oas), resulting in us being able to heavily clean up and standardize how we're dealing with these schemas.
* [x] 🧪 Fleshes out how we test encoded query parameters a bit to match how we're testing this case in [@readme/oas-to-har](https://npm.im/@readme/oas-to-har).
* [x] 🧹 Cleaned up how we handle mocks in tests that have them to make it clearer what's being asserted and not just that the test didn't fail because `nock` caught the expected request.

## 🧬 Testing

See tests.


packages/api/src/index.js
```diff
@@ -1,17 +1,13 @@
...
 const pkg = require('../package.json');
+const { FormDataEncoder } = require('form-data-encoder');
 
 const Cache = require('./cache');
 const { parseResponse, prepareAuth, prepareParams, prepareServer } = require('./lib');
 
-global.fetch = fetch;
-global.Request = fetch.Request;
-global.Headers = fetch.Headers;
-global.FormData = require('form-data');
```
erunion, line 13: We don't need this line anymore because `fetch-har` automatically polyfills `FormData` out for us with the (spec-compliant) [formdata-node](https://npm.im/formdata-node) package now if it's missing. https://github.com/readmeio/fetch-har/blob/main/index.js#L26-L35

packages/api/src/lib/prepareParams.js
```diff
@@ -27,6 +24,61 @@ function isEmpty(obj) {
...
+    if (!fs.existsSync(resolvedFile)) {
+      // It's less than ideal for us to handle files that don't exist like this but because `file`
+      // is a string, it might actually be the full text contents of the file and not actually a
+      // path. Maybe we should regex here and make an attempt to determine if this is a file path
+      // and return an error if it is?
+      // @todo
+      return new Promise(resolve => {
+        resolve(undefined);
+      });
+    }
```
erunion, line 40: I don't love this but I'm also currently unsure how to handle it. On one hand we can regex `file` to see if it *looks* like a file path and isn't just a string, but since this can be a relative path one might assume you just pass in `owlbert.png` instead of `./owlbert.png` and it should just work as `path.resolve()` handles both:

```
> path.resolve('owlbert.png')
'/Users/erunion/code/readmeio/api/owlbert.png'
> path.resolve('./owlbert.png')
'/Users/erunion/code/readmeio/api/owlbert.png'
```

And currently with this resolving to `undefined` that means if you pass in `./owlbert.png` and that file doesn't exist we'll send `./owlbert.png` as whatever parameter that's part of in your API request... which isn't really ideal either.

packages/api/src/lib/prepareParams.js
```diff
@@ -83,68 +135,60 @@ module.exports = async (operation, body, metadata) => {
...
+                return;
+              }
+
+              if (fileMetadata.paramName) {
+                params.body[fileMetadata.paramName] = fileMetadata.base64;
+              } else {
+                params.body = fileMetadata.base64;
+              }
+
+              params.files[fileMetadata.filename] = fileMetadata.buffer;
```
erunion, line 187: With this we'll now be sending file buffers into `fetch-har` to use with the fetch request (instead of relying on the base64).

packages/api/src/index.js
```diff
@@ -56,7 +52,11 @@ class Sdk {
 
         const har = oasToHar(spec, operation, data, prepareAuth(authKeys, operation));
 
-        return fetchHar(har, self.userAgent).then(res => {
+        return fetchHar(har, {
+          userAgent: self.userAgent,
+          files: data.files || {},
```
domharrington, line 57: Would be interesting to think about how we handle file streams in future: https://nodejs.org/docs/latest/api/fs.html#filehandlecreatereadstreamoptions

I'm also not sure what [this](https://nodejs.org/docs/latest/api/fs.html#filehandlereadablewebstream) is, but it seems like a way to create a ReadableStream that's compatible with the ones that you may find in a browser: https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/ReadableStream

TIL: https://nodejs.org/docs/latest/api/webstreams.html#overview

packages/api/src/lib/prepareParams.js
```diff
@@ -27,6 +24,61 @@ function isEmpty(obj) {
   return [Object, Array].includes((obj || {}).constructor) && !Object.entries(obj || {}).length;
 }
 
+function processFile(paramName, file) {
+  if (typeof file === 'string') {
+    // In order to support relative pathed files, we need to attempt to resolve them.
+    const resolvedFile = path.resolve(file);
+    if (!fs.existsSync(resolvedFile)) {
```
domharrington, line 31: Maybe this should be awaited instead of blocking using existsSync? Though I dont actually know what the difference is between `existsSync()` and `await exists()`

domharrington (commented):
This seems great so far! Excited to see how it progresses.

packages/api/src/lib/prepareParams.js
erunion: Yeah good catch, I also don't need to wrap `getStream` in a promise as well because it's already one.

packages/api/src/index.js
erunion: Yep we support `fs.createReadStream` right now. I'm not sure about `fs.createReadStream`. We'll see!

erunion: `fs.createReadStream` is Node 16+ only so we can't easily test against it right now.

packages/api/__tests__/cache.test.js
```diff
@@ -2,9 +2,8 @@ const nock = require('nock');
 const path = require('path');
 const { vol } = require('memfs');
 
-const realFs = jest.requireActual('fs').promises;
+const realFs = jest.requireActual('fs/promises');
```
erunion, line 5: I dropped support for Node 12 in #382 so we can access the promises library this way now.

packages/api/__tests__/integration.test.js
```diff
@@ -0,0 +1,150 @@
...
+        'User-Agent': expect.stringMatching(/api \(node\)\/\d+.\d+.\d+/),
+      }),
+      json: null,
+      method: 'POST',
+      origin: expect.any(String),
+      url: 'https://httpbin.org/anything/v1/oa_citations/records',
+    });
+  });
+
+  // it.skip('should send along required parameters if not supplied', async () => {
```
erunion, line 37: We don't support sending required props that have defaults right now. Ticketed to RM-3369.

packages/api/__tests__/integration.test.js
```diff
@@ -0,0 +1,150 @@
...
+        'User-Agent': expect.stringMatching(/api \(node\)\/\d+.\d+.\d+/),
+      }),
+      json: null,
+      method: 'POST',
+      origin: expect.any(String),
+      url: 'https://httpbin.org/anything/multipart-formdata',
+    });
+  });
+
+  // it.only('should support `multipart/form-data` requests with an array of files', async () => {
```
erunion, line 140: This doesn't send anything right now and is a larger problem with the `prepareParams` library that I plan on resolving in another PR.

packages/api/__tests__/integration.test.js
```diff
@@ -0,0 +1,158 @@
...
+const fileUploads = require('@readme/oas-examples/3.0/json/file-uploads.json');
+
+beforeEach(() => {
+  // Though this test doesn't mock anything out because nock is used within every other test in
+  // this suite there's an issue where if you run them all together nock will be enabled here and
+  // any request that happens will fail because nock reports them as being unmocked.
+  nock.restore();
+});
+
+describe.skip('`application/x-www-form-urlencoded`', () => {
```
erunion, line 13: All of these tests in this file work fine locally but when they're run in CI `nock` from other tests are carrying through over here and every test fails because nock doesn't think that they're been mocked out.

Because I plan on creating a browser-ready version of `api` as part of the install process, and will be migrating tests over to Mocha + Chai then so we can use the same tests for it in Node and with Karma I'm going to leave these skipped for now until that migration happens, at which point they'll be able to be run again without any troubles.

packages/api/__tests__/integration.test.js
erunion: This test also uncovered a problem with `prepareParams` where the metadata that we're sending here for the path is also being sent in the form payload:

```json
{
  args: {},
  data: '',
  files: {},
  form: { dataset: 'v1', version: 'oa_citations' },
  headers: {
    Accept: 'application/json',
    'Accept-Encoding': 'gzip,deflate',
    'Content-Length': '31',
    'Content-Type': 'application/x-www-form-urlencoded',
    Host: 'httpbin.org',
    'User-Agent': 'api (node)/4.2.0',
    'X-Amzn-Trace-Id': 'REDACTED'
  },
  json: null,
  method: 'POST',
  origin: 'REDACTED',
  url: 'https://httpbin.org/anything/v1/oa_citations/records'
}
```

packages/api/__tests__/integration.test.js
```diff
@@ -0,0 +1,150 @@
...
+      orderId: 1234,
+      userId: 5678,
+      documentFile: `${__dirname}/__fixtures__/hello.txt`,
+    };
+
+    await expect(api(fileUploads).post('/anything/multipart-formdata', body)).resolves.toStrictEqual({
+      args: {},
+      data: '',
+      files: {
+        documentFile: `Hello world!
```
Dashron, line 125: should we do any testing of unicode or other encoding options?

packages/api/src/lib/prepareParams.js
```diff
@@ -27,6 +24,64 @@ function isEmpty(obj) {
   return [Object, Array].includes((obj || {}).constructor) && !Object.entries(obj || {}).length;
 }
 
+function processFile(paramName, file) {
```
Dashron, line 27: this could use a comment describing roughly what it's doing, and how.

Dashron (changes requested):
there's a *lot* of dense technical work here, so it's hard for me to be super confident but the tests look good. left a couple small comments.

packages/api/__tests__/integration.test.js
erunion: ill add a unicode example

packages/api/src/lib/prepareParams.js
erunion: I'll do you on better in a different PR and load in the JSDoc ESLint plugin that we have in `oas` and add comments to everything.

Dashron (approved):

PR #389, opened 2022-02-03
fix: issue where formData would sometimes be sent twice

erunion:
## 🧰 Changes

This resolves a bug I discovered in #380 where if you sent **only** a metadata object to an `x-www-form-urlencoded` opration we'd copy that metadata over to being the `formData` object, ultimately potentially things like path parameters in the body payload.

## 🧬 QA & Testing

See the test I added for this and the test I updated that had this incorrect behavior as an expectation. Oops.


packages/api/src/lib/prepareParams.js
```diff
@@ -131,6 +132,7 @@ module.exports = async (operation, body, metadata) => {
         // as a metadata object and organize into parameters.
         // eslint-disable-next-line no-param-reassign
         metadata = body;
+        metadataIntersected = true;
```
erunion, line 135: Open to suggestions on a better variable name than this. Basically I just need to know when `body` was copied to `metadata`, where what the user sent was a metadata object but since they didn't send a body, and `body` is the first argument to an operation we need to rewrite it a bit.

It's confusing I know and I'm not explaining it properly.

Issue #38, opened 2020-05-21
Add support for cookie parameters


erunion:
This'll require updates to here and also [@readme/oas-to-har](https://www.npmjs.com/package/@readme/oas-to-har).

erunion:
In addition to adding support for cookie params, we might as well start to factor in the case where a parameter name might be shared across multiple inputs (query, header, cookie). We should adopt a reserved keyword syntax in the `metadata` object for these as such:

```js
sdk.post('/har', {
  foo: ['bar', 'baz'], 
  baz: 'abc',
  key: 'value',
  $cookies: {
    foo: 'bar',
    bar: 'baz'
  }
})
```

erunion:
Cookie support is now available in `@readme/oas-to-har` so we just need to pull in support for it into here and add the `$cookies` override.

erunion:
Completed in #393

PR #393, opened 2022-02-07
feat: adding support for cookie parameters

erunion:
| 🚥  | Fix RM-3242 |
| :-- | :--- |

## 🧰 Changes

Adds full support for cookie parameters in SDK requests.

Cookie support was already built into `httpsnippet-client-api`, but we only supported cookies in auth within the SDK.

## 🧬 QA & Testing

See tests.

### DIFF
--- a/packages/api/__tests__/lib/prepareParams.test.ts
+++ b/packages/api/__tests__/lib/prepareParams.test.ts
@@ -323,5 +323,49 @@ describe('#prepareParams', () => {
     });
   });
 
-  it.todo(`should be able to handle parameters when they're defined as common parameters`);
+  describe('parameter types', () => {
+    let parameterStyle;
+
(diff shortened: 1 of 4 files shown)


packages/api/__tests__/lib/prepareParams.test.ts
```diff
@@ -323,5 +323,49 @@ describe('#prepareParams', () => {
     });
   });
 
-  it.todo(`should be able to handle parameters when they're defined as common parameters`);
+  describe('parameter types', () => {
```
erunion, line 326: `body`, `files`, and `formData` are being tested elsewhere.

kanadgupta:
You should be able to close out https://github.com/readmeio/api/issues/38 now!

PR #396, opened 2022-02-08
feat: overhauling how we process parameters to support required defaults

erunion:
| 🚥  | Fix RM-3369 |
| :-- | :--- |

## 🧰 Changes

This **completely** overhauls how we process and ingest request body payloads and parameter metadata so that we now augment any supplied data with any required request body payload property or parameter that has a documented default.

For example in the [USPTO example spec](https://github.com/readmeio/oas-examples/blob/main/3.0/json/uspto.json) it has a  `POST /{dataset}/{version}/records` operation with a `dataset` and `version` path parameter and an object payload. When you run this currently **without** sending any body or metadata we'll make an API call to `/{dataset}/{version}/records` with an empty payload.

With these fixes, since `dataset` and `version` are required parameters, and have documented defaults, and because in its object payload there is a `criteria` property with a default of `*:*` now when you make an API call with `sdk.post('/{dataset}/{version}/records)` we'll hit `/oa_citations/v1/records` with a payload of `{ criteria: '*:*'}` all without the user having to do anything.

It is very neat and works pretty well!

This functionality also mostly functions identicaly to how we handle form defaults in the reference guide, I'm just using a different library for traversing our JSON Schema and composing defaults.

## 🧬 QA & Testing

See the new tests I wrote or you can run this code to see it in action:

```js
const spec = require('@readme/oas-examples/3.0/json/uspto.json');
spec.servers[0].url = '{scheme}://httpbin.org/anything';

const sdk = require('.')(spec);

sdk.post('/{dataset}/{version}/records').then(res => {
  console.log(res);
});

/*
{
  args: {},
  data: '',
  files: {},
  form: { criteria: '*:*' },
  headers: {
    Accept: 'application/json',
    'Accept-Encoding': 'gzip,deflate',
    'Content-Length': '14',
    'Content-Type': 'application/x-www-form-urlencoded',
    Host: 'httpbin.org',
    'User-Agent': 'api (node)/4.2.0',
    'X-Amzn-Trace-Id': '<<REDACTED>>'
  },
  json: null,
  method: 'POST',
  origin: '<<REDACTED>>',
  url: 'https://httpbin.org/anything/oa_citations/v1/records'
}
*/
```


packages/api/src/lib/prepareParams.ts
```diff
@@ -28,7 +38,40 @@ function isEmpty(obj: any) {
...
+function isPrimitive(obj: any) {
+  return typeof obj === null || typeof obj === 'number' || typeof obj === 'string';
+}
+
+function merge(src: any, target: any) {
+  if (Array.isArray(target)) {
+    // @todo we need to add support for merging array defaults with array body/metadata arguments
+    return target;
+  } else if (!isObject(target)) {
+    return target;
```
erunion, line 58: Because we let you supply primitive values or a readable stream as a body argument we can't merge that with anything so we're dumping out and just returning whatever the user supplied.

packages/api/src/lib/prepareParams.ts
```diff
@@ -105,165 +165,171 @@ export default async function prepareParams(operation: Operation, body?: unknown
...
-      if (intersection && intersection / Object.keys(body).length > 0.25) {
-        // If more than 25% of the body intersects with the parameters that we've got on hand, then
-        // we should treat it as a metadata object and organize into parameters.
-        // eslint-disable-next-line no-param-reassign
-        metadata = body as Record<string, unknown>; // @todo we should just do a "if body isnt an object don't digest it"
-        metadataIntersected = true;
+  // If a body argument was supplied we need to do a bit of work to see if it's actually a body
+  // argument or metadata because the library lets you supply either a body, metadata, or body with
+  // metadata.
+  if (typeof body !== 'undefined') {
```
erunion, line 173: I hope this code is a lot cleaner for you all to read and understand then what it was.

erunion:
@readmeio/api-tooling I'm merging this in because I'm already in merge hell with the changes here and #399 so if you manage to take a look at this this week please still leave feedback and I'll address them in different PRs.

PR #397, opened 2022-02-09
fix: fixing how we generate the dist to work under CJS imports

erunion:
| 🚥  | Fix RM-3394 |
| :-- | :--- |

## 🧰 Changes

This rolls back the `target: es2020` config I added to our TS config when I migrated the library to TS as I didn't really understand the implications of what was going on with it.

The reason I added it in the first place was this error:

screen_shot_2022-02-08_at_2 49 10_pm

Because `form-data-encoder` uses `#private` class member declarations TS can only compile that to ES2015+ targets. Thankfully with TS 4.6 they're adding a new `node12` module and moduleResolution config that's able to ignore these at the expense of using types elsewhere. https://www.typescriptlang.org/tsconfig#node12nodenext-nightly-builds

And because `form-data-encoder` doesn't have a `@types/form-data-encoder` package we need to declare a module for it ourselves, resulting in type loss.

It's an unfortunate loss but because IMHO it's still far too early to make libraries ESM-only I don't want to migrate `api` towards that direction yet. Maybe when Node 14 hits EOL but we'll see.

## 🧬 QA & Testing

It's hard to test how this was broken without manually reverting the TS config changes but the dist unit test I wrote loads the library without a `{default: api}` import, and running the `example.js` script also, works so if this test runs in CI it should be good?

### DIFF
--- a/packages/api/__tests__/tsconfig.json
+++ b/packages/api/__tests__/tsconfig.json
@@ -2,6 +2,7 @@
   "extends": "../tsconfig.json",
   "compilerOptions": {
     "module": "es2020",
+    "moduleResolution": "node",
     "noImplicitAny": false,
     "resolveJsonModule": true
   },
--- a/packages/api/__tests__/tsconfig.json
+++ b/packages/api/__tests__/tsconfig.json
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "module": "es2020",
     "noImplicitAny": false,
+    "resolveJsonModule": true
   },
   "include": ["../src/**/*", "*.ts", "**/*"]
 }
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -7,7 +7,8 @@
   "scripts": {
     "build": "tsc",
     "lint": "eslint . --ext .js,.ts",
-    "prebuild": "rm -rf dist/ @types/",
+    "prebuild": "rm -rf dist/ @types/; npm run prebuild.packageConfig",
+    "prebuild.packageConfig": "node -p \"'export const PACKAGE_NAME = \\'' + require('./package.json').name + '\\';\\nexport const PACKAGE_VERSION = \\'' + require('./package.json').version + '\\';'\" > src/package.ts",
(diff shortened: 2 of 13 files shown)


packages/api/package.json
```diff
@@ -7,7 +7,8 @@
   "scripts": {
     "build": "tsc",
     "lint": "eslint . --ext .js,.ts",
-    "prebuild": "rm -rf dist/ @types/",
+    "prebuild": "rm -rf dist/ @types/; npm run prebuild.packageConfig",
+    "prebuild.packageConfig": "node -p \"'export const PACKAGE_NAME = \\'' + require('./package.json').name + '\\';\\nexport const PACKAGE_VERSION = \\'' + require('./package.json').version + '\\';'\" > src/package.ts",
```
erunion, line 11: Had to add this build step for our package file because we were loading `../package.json` in a few files and TS was transpiling the dist directory into the following:

screen_shot_2022-02-08_at_5 58 08_pm

packages/api/src/index.ts
```diff
@@ -201,6 +201,9 @@ class Sdk {
   }
 }
 
-export default function api(uri: string | OASDocument) {
+// Why `export` vs `export default`? If we leave this as `export` then TS will transpile it into
+// a `module.exports` export so that when folks load this they don't need to load it as
+// `require('api').default`.
+export = (uri: string | OASDocument) => {
```
erunion, line 207: I want to preserve doing `require('api')(spec)` as long as we can.

packages/api/__tests__/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "module": "es2020",
     "noImplicitAny": false,
+    "resolveJsonModule": true
```
erunion, line 6: The dist test imports a JSON spec so this needed to be added.

erunion:
Tests currently do not pass because the module resoltuion stuff from the root is carrying through to tests where it doesn't need to be applied. Working on it.

PR #404, opened 2022-02-25
feat: move fetching logic within api out into core

erunion:
## 🧰 Changes

In preparation of the codegenerator I'm moving the core fetching logic out of `api` and into the core library so we can re-use this logic within autogenerated SDKs.

This ultimately makes the existing `api` package and its next-level Proxy work a thin wrapper on top of a more structured system for making `api`-like requests to an API.

## 🧬 QA & Testing

All tests and builds still passing?


packages/api/test/dist.test.ts
```diff
@@ -26,4 +27,22 @@ describe('typescript dist verification', function () {
 
     mock.done();
   });
+
+  it('should be able to set an auth token', async function () {
```
erunion, line 31: Added this new test because the `.auth()`, `.config()` and `.server()` accessors don't talk to the specs cache so I needed to make sure that the new APICore class was being handled properly there in the dist.

packages/core/test/index.test.ts
```diff
@@ -0,0 +1,302 @@
+import { assert, expect } from 'chai';
```
erunion, line 1: Most of these tests are all similar to tests that are happening in the `api/` package, just without the proxy and SDK chaining interface.

PR #411, opened 2022-03-09
feat: TypeScript SDK code generation 🧙 

erunion:
| 🚥 Fix RM-3414 |
| :-- |

## 🧰 Changes

This implements a first pass on TypeScript code generation of a functional `api` SDK by heavily utilizing the amazing [ts-morph](https://ts-morph.com/) library. Generated SDKs with this code here are fully TS ready and have exported types, [method overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads), auto-docblocks, `api`-like operation accessors, everything. 

If it's in `api`, it's here.

## 🧬 QA & Testing

Really difficult to test this all right now as it's not hooked up to anything, and the TS that is being exported here doesn't work because `api/core` doesn't exist in our currently on NPM `api` package, but an **extremely thorough** test suite is incoming.

That said, this is an example of what one of these SDKs looks like:

```ts
import Oas from 'oas';
import APICore from './src/core';
import definition from './test/__fixtures__/simple.oas.json';

export default class SDK {
  spec: Oas;
  core: APICore;
  authKeys: (number | string)[][] = [];

  constructor() {
    this.spec = Oas.init(definition);
    this.core = new APICore(this.spec, 'api/1.0.0');
  }

  /**
   * Optionally configure various options, such as response parsing, that the SDK allows.
   *
   * @param config Object of supported SDK options and toggles.
   * @param config.parseResponse If responses are parsed according to its `Content-Type` header`.
   */
  config(config: ConfigOptions) {
    this.core.setConfig(config);
  }

  /**
   * If the API you're using requires authentication you can supply the required credentials
   * through this method and the library will magically determine how they should be used
   * within your API request.
   *
   * With the exception of OpenID and MutualTLS, it supports all forms of authentication
   * supported by the OpenAPI specification.
   *
   * @example <caption>HTTP Basic auth</caption>
   * sdk.auth('username', 'password');')
   *
   * @example <caption>Bearer tokens (HTTP or OAuth 2)</caption>
   * sdk.auth('myBearerToken');
   *
   * @example <caption>API Keys</caption>
   * sdk.auth('myApiKey');
   *
   * @see {@link https://spec.openapis.org/oas/v3.0.3#fixed-fields-22}
   * @see {@link https://spec.openapis.org/oas/v3.1.0#fixed-fields-22}
   * @param values Your auth credentials for the API; can specify up to two strings or numbers.
   */
  auth(...values: string[] | number[]) {
    this.core.setAuth(...values);
    return this;
  }

  /**
   * If the API you're using offers alternate server URLs, and server variables, you can tell
   * the SDK which one to use with this method. To use it you can supply either one of the
   * server URLs that are contained within the OpenAPI definition (along with any server
   * variables), or you can pass it a fully qualified URL to use (that may or may not exist
   * within the OpenAPI definition).
   *
   * @example <caption>Server URL with server variables</caption>
   * sdk.server('https://{region}.api.example.com/{basePath}', {
   *   name: 'eu',
   *   basePath: 'v14',
   * });
   *
   * @example <caption>Fully qualified server URL</caption>
   * sdk.server('https://eu.api.example.com/v14');
   *
   * @param url Server URL
   * @param variables An object of variables to replace into the server URL.
   */
  server(url: string, variables = {}) {
    this.core.setServer(url, variables);
  }

  /**
   * Multiple status values can be provided with comma separated strings
   *
   * @param metadata Object containing all path, query, header, and cookie parameters to supply.
   */
  get(path: string, metadata: FindPetsByStatusMetadataParam): Promise<FindPetsByStatus_Response_200>;
  get(path: string, metadata?: Record<string, unknown>): Promise<unknown> {
    return this.core.fetch(path, 'get', metadata);
  }

  /**
   * Multiple status values can be provided with comma separated strings
   *
   * @param metadata Object containing all path, query, header, and cookie parameters to supply.
   */
  findPetsByStatus(metadata: FindPetsByStatusMetadataParam): Promise<FindPetsByStatus_Response_200> {
    return this.core.fetch('/pet/findByStatus', 'get', metadata);
  }
}

interface ConfigOptions {
  /**
   * By default we parse the response based on the `Content-Type` header of the request. You
   * can disable this functionality by negating this option.
   */
  parseResponse: boolean;
}
export type FindPetsByStatusMetadataParam = {
  /**
   * Status values that need to be considered for filter
   */
  status: ("available" | "pending" | "sold")[];
  [k: string]: unknown;
};
export type FindPetsByStatus_Response_200 = Pet[];
export interface Pet {
  id?: number;
  category?: Category;
  name: string;
  photoUrls: string[];
  tags?: Tag[];
  /**
   * pet status in the store
   */
  status?: "available" | "pending" | "sold";
  [k: string]: unknown;
}
export interface Category {
  id?: number;
  name?: string;
  [k: string]: unknown;
}
export interface Tag {
  id?: number;
  name?: string;
  [k: string]: unknown;
}
```

### DIFF
--- a/packages/api/src/codegen/typescript.ts
+++ b/packages/api/src/codegen/typescript.ts
@@ -1,0 +1,355 @@
+import type Oas from 'oas';
+import type { Operation } from 'oas';
+import type { HttpMethods, JSONSchema } from 'oas/@types/rmoas.types';
+import type { MethodDeclaration } from 'ts-morph';
+
+import CodeGenerator from '.';
+import { IndentationText, Project, QuoteKind } from 'ts-morph';
+import { compile } from 'json-schema-to-typescript';
+
+type OperationTypeHousing = {
+  types: {
+    params: Record<'body' | 'formData' | 'metadata', string>;
+    responses?: Record<string, string>;
+  };
+  operation: Operation;
+};
+
+export default class TSGenerator extends CodeGenerator {
+  userAgent: 'api/1.0.0';
+
+  project: Project;
+
+  types: Map<string, string>;
+
+  files: Record<string, string>;
+
+  constructor(spec: Oas) {
+    super(spec);
+
+    // @todo fill this user agent in with something contextual.
+    this.userAgent = 'api/1.0.0';
+
+    this.project = new Project({
+      manipulationSettings: {
+        indentationText: IndentationText.TwoSpaces,
+        quoteKind: QuoteKind.Single,
+      },
+      compilerOptions: {
+        // target: ScriptTarget.ES3,
+        outDir: 'dist',
+      },
+    });
+
+    this.types = new Map();
+  }
+
+  /**
+   * Compile the current OpenAPI definition into a TypeScript library.
+   *
+   */
+  async generator() {
+    const { operations, methods } = await this.loadOperationsAndMethods();
+
+    const sdkSource = this.project.createSourceFile('index.ts', 'export default class SDK {}');
+
+    sdkSource.addImportDeclarations([
+      { defaultImport: 'Oas', moduleSpecifier: 'oas' },
+      { defaultImport: 'APICore', moduleSpecifier: 'api/core' },
+      {
+        defaultImport: 'definition',
+        moduleSpecifier: '../fixtures/simple.oas.json', // './openapi.json',
+      },
+    ]);
+
+    const sdk = sdkSource.getClassOrThrow('SDK');
+    sdk.addProperty({ name: 'spec', type: 'Oas' });
+    sdk.addProperty({ name: 'core', type: 'APICore' });
+    sdk.addProperty({ name: 'authKeys', type: '(number | string)[][]', initializer: '[]' });
+
+    sdk.addConstructor({
+      statements: ['this.spec = Oas.init(definition);', `this.core = new APICore(this.spec, '${this.userAgent}');`],
+    });
+
+    // Add our core API methods for controlling auth, servers, and various configurable abilities.
+    sdk.addMethods([
+      {
+        name: 'config',
+        parameters: [{ name: 'config', type: 'ConfigOptions' }],
+        statements: 'this.core.setConfig(config);',
+      },
+      {
+        name: 'auth',
+        parameters: [{ name: '...values', type: 'string[] | number[]' }],
+        statements: ['this.core.setAuth(...values);', 'return this;'],
+      },
+      {
+        name: 'server',
+        parameters: [
+          { name: 'url', type: 'string' },
+          { name: 'variables', initializer: '{}' },
+        ],
+        statements: 'this.core.setServer(url, variables);',
+      },
+    ]);
+
+    sdkSource.addInterface({
+      name: 'ConfigOptions',
+      properties: [
+        {
+          name: 'parseResponse',
+          type: 'boolean',
+        },
+      ],
+    });
+
+    // Add all common method accessors into the SDK.
+    const methodGenerics: Record<string, MethodDeclaration> = {};
+    methods.forEach(method => {
+      const fetchArgs = ['path', `'${method}'`];
+      const parameters = [{ name: 'path', type: 'string' }];
+
+      // Method generic body + metadata parameters are always optional.
+      if (method !== 'get') {
+        parameters.push({ name: 'body?', type: 'unknown' });
+        fetchArgs.push('body');
+      }
+
+      parameters.push({ name: 'metadata?', type: 'Record<string, unknown>' });
+      fetchArgs.push('metadata');
+
+      methodGenerics[method] = sdk.addMethod({
+        name: method,
+        returnType: 'Promise<unknown>',
+        parameters,
+        statements: `return this.core.fetch(${fetchArgs.join(', ')});`,
+      });
+    });
+
+    // Add all available operation ID accessors into the SDK.
+    Object.entries(operations).forEach(([operationId, data]: [string, OperationTypeHousing]) => {
+      const operation: Operation = data.operation;
+      const fetchArgs = [`'${operation.path}'`, `'${operation.method}'`];
+      const parameters = [];
+      if (data.types.params) {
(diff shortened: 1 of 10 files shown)


packages/api/src/codegen/typescript.ts
```diff
@@ -0,0 +1,357 @@
...
+
+    // @todo fill this user agent in with something contextual.
+    this.userAgent = 'api/1.0.0';
+
+    this.project = new Project({
+      manipulationSettings: {
+        indentationText: IndentationText.TwoSpaces,
+        quoteKind: QuoteKind.Single,
+      },
+      compilerOptions: {
```
erunion, line 38: These compiler options will eventually be controlled via the forthcoming cli and abide by how folks want their autogen'd SDKs to target and output to.

packages/api/src/codegen/typescript.ts
```diff
@@ -0,0 +1,357 @@
...
+    const { operations, methods } = await this.loadOperationsAndMethods();
+
+    const sdkSource = this.project.createSourceFile('index.ts', 'export default class SDK {}');
+
+    sdkSource.addImportDeclarations([
+      { defaultImport: 'Oas', moduleSpecifier: 'oas' },
+      { defaultImport: 'APICore', moduleSpecifier: 'api/core' },
+      {
+        defaultImport: 'definition',
+        moduleSpecifier: '../fixtures/simple.oas.json', // './openapi.json',
```
erunion, line 61: This'll get swapped out with the real OAS path eventually as I write a test suite for all of this.

packages/api/src/codegen/typescript.ts
```diff
@@ -0,0 +1,357 @@
...
+   */
+  async convertJSONSchemaToTypescript(schema: JSONSchema, name: string) {
+    // Though our JSON Schema type exposes JSONSchema4, which `json-schema-to-typescript`, it won't
+    // accept our custom type.
+    const ts = await compile(schema as any, name, {
+      bannerComment: '',
+    });
+
+    let primaryType: string;
+    const tmp = this.project.createSourceFile(`${name}.types.tmp.ts`, ts);
```
erunion, line 216: Because [json-schema-to-typescript](https://npm.im/json-schema-to-typescript) outputs only a string containing our exported TS types we need to do a bit of post-processing on it in order to extract a few things:

* The primary type that this JSON Schema object was created for (as dictated by the `name` argument). So if we're converting the `metadata` parameter JSON Schema for a `findByPets` operation, and `name` in this case is `FindByPetsMetadata`, then we'll be looking the primary type of `FindByPetsMetadata` so we can add it into the `findByPets()` operation accessor.
* All types and interfaces that were created with this JSON Schema object. Because we'll need to actually insert these types and interfaces into our generated SDK we need to make sure that we don't duplicate them so we need to do some more post-processing here to put them into a set of the exported name to the exported TS type string content.

If this all sounds like a lot to mentally process it's because it is.

This code also isn't optimized either because we're going to definitely be processing the same JSON Schema object sometimes. A problem for another day.

packages/api/src/codegen/typescript.ts
```diff
@@ -0,0 +1,357 @@
...
+      if (!primaryType) {
+        primaryType = declarationName;
+      }
+
+      declarations.get(declarationName).forEach(declaration => {
+        this.types.set(declarationName, declaration.getText());
+      });
+    });
+
+    this.project.removeSourceFile(tmp);
```
erunion, line 229: Since we're only temporarily using `ts-morph` to do this post-processing on the `json-schema-to-typescript` types and we don't want to put those types into their own files (we'll do that later), we're immediately removing our temporary source file from the overall autogenerated SDK source code.

packages/api/src/codegen/typescript.ts
```diff
@@ -0,0 +1,357 @@
...
+   * along with every HTTP method that's in use.
+   *
+   */
+  async loadOperationsAndMethods() {
+    const operations: Record</* operationId */ string, OperationTypeHousing> = {};
+    const methods: string[] = [];
+
+    await Promise.all(
+      Object.entries(this.spec.getPaths()).map(async ([, ops]) => {
+        await Promise.all(
```
erunion, line 248: This nested promise work is pretty nasty but because `getParameterTypesForOperation` and `getResponseTypeosForOperation` are async it needs to happen (unless of course you know of a better way to clean this up).

packages/api/src/codegen/typescript.ts
```diff
@@ -0,0 +1,357 @@
...
+   * usable TypeScript types.
+   *
+   * @param operation
+   * @param operationId
+   */
+  async getParameterTypesForOperation(operation: Operation, operationId: string) {
+    let params;
+    const paramSchemas = operation.getParametersAsJsonSchema({
+      mergeIntoBodyAndMetadata: true,
+      retainDeprecatedProperties: true,
```
erunion, line 286: These two options were added in https://github.com/readmeio/oas/pull/617 to allow me to merge all metadata-categorized parameters (query, header, cookie, path) into a single `metadata` JSON Schema entry -- and ultimately a single TS type for that parameter.

Dashron:
the comment on the following looks wrong

```
  /**
   * Multiple status values can be provided with comma separated strings
   *
   * @param metadata Object containing all path, query, header, and cookie parameters to supply.
   */
  get(path: string, metadata: FindPetsByStatusMetadataParam): Promise<FindPetsByStatus_Response_200>;
  get(path: string, metadata?: Record<string, unknown>): Promise<unknown> {
    return this.core.fetch(path, 'get', metadata);
  }
```

also you could use generics if you want to tie the two `unknown` values together. e.g.
```get<T>(path: string, metadata?: Record<string, T>): Promise<T> {```

This would mean that it would know the return type based on introspection of the metadata type. Not sure if this is intended though, you might just want two separate unknowns!

erunion:
Talked offline with @Dashron and I'm going to do the generic changes, but I'm doing them as part of the work happening in #413.

Dashron (approved):

PR #413, opened 2022-03-10
test: SDK codegeneration test suite

erunion:
| 🚥 Fix RM-3414 |
| :-- |

## 🧰 Changes

This creates a new unit test suite with a `toMatchSDKFixture` Chai plugin that allows us to store real SDK code generation within the repository and manage them as real files, versus Jest-like snapshots.

In the process of writing these tests I've also fixed a heap of bugs and quirks:

* [x] 🆕 If a request body is optional, but there's metadata present, then we'll create a couple method overloads for the operation and HTTP method generics. The operation accessor will both get an overload with `body` and `metadata` as required parameters, another with only a required `metadata` parameter, and the original operation accessor will have both `body` and `metadata` as optional. Rinse, repeat for the HTTP method generics.
* [x] 🆕 If a response doesn't have a defined return type (read: schema), then we'll now type the method with a `<T = unknown>` generic, along with a typed `Promise<T>` return type, so that the user can type the response however they wish. This does **not** apply if the operation has a defined response schema (read: type).
* [x] 🔧 Changed how we generate docblocks for our method overloads so that now if a parameter has a known type, and isn't `any` or `Record<string, unknown>`, then it won't get a `@param` docblock. This'll allow us to avoid needless and duplicate boilerplate noise throughout the entire SDK file on every accessor. If someone is using an IDE that supports TS types they'll have much better docs anyways with that over generalized `@param metadata This is an object of parameters!"` nonsense.
* [x] 🐛 Fixed a bug where we were generating a generic method accessor *n* times for every operation in the OAS definition.
  * For example, the Petstore has 7 `POST` operations so we were create 7 HTTP method generics along with 7 overloads on those generics for each operation instead of a single method generic + 7 overloads. Oops!
* [x] 🐛 Fixed an issue where if at type had a string enum in it we'd escape the quotes in the type when adding them into the SDK, resulting in invalid code.
* [x] ⚡ Moved the Prettier call that happens within [json-schema-to-typescript](https://npm.im/json-schema-to-typescript) into a single call on the entire source file after we're done generating it as running it once for every type we'd generate was so, so, so slow.
  * [ts-morph](https://npm.im/ts-morph) already mostly handles this well enough with its [manipulation settings](https://ts-morph.com/manipulation/) but because we're generating types outside of it and inserting them into the SDK as string content they don't get formatted by `ts-morph`. Unfortunately this change has required me to make Prettier as a library dependency.
 
I've also made a futile attempt to speed things along further by memoizing the [json-schema-to-typescript](https://npm.im/json-schema-to-typescript) call but because of the way these calls are happening inside of a `Promise.all()` they don't always get memoized and we end up converting the commonly used schemas over and over. When I make a performance pass this will be the first thing I plan on tackling because right now... uh

Screen Shot 2022-03-17 at 5 37 50 PM

## 🧬 QA & Testing

There's no way yet to generate an SDK for any given OpenAPI definition but with this new fixture system I've written there are some pre-generated fixtures in https://github.com/readmeio/api/tree/test/codegen-testsuite/packages/api/test/__fixtures__/sdk.

* [ ] Tests should pass.
* [ ] Generated SDKs should look good.


packages/api/test/__fixtures__/sdk/petstore/index.ts
```diff
@@ -0,0 +1,592 @@
...
+   * Creates list of users with given input array
+   *
+   * @param body Request body payload data.
+   */
+  post(path: string, body: CreateUsersWithListInputBodyParam): Promise<unknown>;
+  /**
+   * Uploads an image
+   *
+   * @param body Request body payload data.
+   * @param metadata Object containing all path, query, header, and cookie parameters to supply.
```
erunion, line 110: I really don't like these docblocks but I also don't know what else to do aside from maybe dumping the that the accompanying types (like `UploadFileBodyParam` have in here -- and I definitely know I won't like that.

packages/api/test/__fixtures__/sdk/optional-payload/index.ts
```diff
@@ -0,0 +1,148 @@
+import Oas from 'oas';
```
erunion, line 1: This test fixture is to explicitly test how we're handling operations that have required metadata and optional payloads and to make sure that we add the necessary type overloads for them.

packages/api/test/__fixtures__/sdk/petstore/index.ts
```diff
@@ -0,0 +1,617 @@
...
+  /**
+   * Updates a pet in the store with form data
+   *
+   */
+  updatePetWithForm<T = unknown>(metadata: UpdatePetWithFormMetadataParam): Promise<T>;
+  /**
+   * Updates a pet in the store with form data
+   *
+   */
+  updatePetWithForm<T = unknown>(
```
erunion, line 290: I really do not like these duplicate docblocks but I also don't really know how better to handle it. Maybe the docblock could go on the core/non-overloaded method only?

packages/api/test/__fixtures__/sdk/simple/index.ts
```diff
@@ -79,18 +75,22 @@ export default class SDK {
    * Multiple status values can be provided with comma separated strings
    *
    * @summary Finds Pets by status
-   * @param metadata Object containing all path, query, header, and cookie parameters to supply.
```
erunion, line 82: I've changed these parameter docblocks to only be present if the parameter doesn't have a type on it so we can avoid a lot of this duplicate boilerplate everywhere.

packages/api/.eslintrc
```diff
@@ -19,6 +19,7 @@
     // Though we aren't requiring JSDoc blocks to be present, if they are they should be properly
     // formatted.
     "jsdoc/require-jsdoc": "off",
+    "jsdoc/tag-lines": "off",
```
erunion, line 22: Disabling this because I think the `@example` docs look cleaner with a space between them.

Screen Shot 2022-03-17 at 5 56 36 PM

packages/api/src/codegen/typescript.ts
```diff
@@ -1,15 +1,27 @@
...
 import { IndentationText, Project, QuoteKind } from 'ts-morph';
 import { compile } from 'json-schema-to-typescript';
+import { format as prettier } from 'json-schema-to-typescript/dist/src/formatter';
+
+const memoizedCompile = memoize(compile);
 
 type OperationTypeHousing = {
   types: {
-    params: Record<'body' | 'formData' | 'metadata', string>;
+    params?: Record<'body' | 'formData' | 'metadata', string>;
```
gratcliff, line 24: I see you got around the `overload` issue by making them both optional 😅

packages/api/src/codegen/typescript.ts
```diff
@@ -190,160 +202,272 @@ sdk.server('https://eu.api.example.com/v14');`)
...
+        parameters: [{ ...parameters.metadata }],
+        returnType,
+        docs: docblock ? [docblock] : null,
       });
 
-      // Add a typed generic HTTP method overload for this operation.
-      if (operation.method in methodGenerics) {
-        methodGenerics[operation.method].addOverload({
-          parameters: [{ name: 'path', type: 'string' }, ...parameters],
+      // Create an overload that has both `body` and `metadata` parameters as optional. Eventhough
```
gratcliff, line 423:
```suggestion
      // Create an overload that has both `body` and `metadata` parameters as optional. Even though
```

packages/api/src/codegen/typescript.ts
```diff
@@ -190,160 +202,272 @@ sdk.server('https://eu.api.example.com/v14');`)
...
+      // our `metadata` parameter is actually required for this operation this the only way we're
```
gratcliff, line 424:
```suggestion
      // our `metadata` parameter is actually required for this operation this is the only way we're
```

gratcliff (approved):
Looks good, @erunion!

packages/api/src/codegen/typescript.ts
```diff
@@ -503,6 +629,18 @@ sdk.server('https://eu.api.example.com/v14');`)
...
+      .then(res => res.reduce((prev, next) => Object.assign(prev, next), {}))
+      .then(res => (Object.keys(res).length ? res : undefined));
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  formatter(content: string) {
+    return prettier(content, {
+      format: true,
+      style: {
+        printWidth: 120,
```
ilias-t, line 641: do you think it's worth pulling these values from some config?

ilias-t (approved):

packages/api/src/codegen/typescript.ts
```diff
@@ -190,160 +202,272 @@ sdk.server('https://eu.api.example.com/v14');`)
...
+  ) {
+    const docblock: OptionalKind<JSDocStructure> = { tags: [] };
+    const summary = operation.getSummary();
+    const description = operation.getDescription();
+    if (summary || description) {
+      // To keep our generated docblocks clean we should only add the `@summary` tag if we've
+      // got both a summary and a description present on the operation, otherwise we can alternate
+      // what we surface the main docblock description.
+      docblock.description = writer => {
+        if ((description && summary) || (description && !summary)) {
```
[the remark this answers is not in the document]
erunion: uh. yeah. i dont know why i made this block so goddamned complicated.

packages/api/test/__fixtures__/sdk/petstore/index.ts
```diff
@@ -0,0 +1,617 @@
...
+   *
+   * @param config Object of supported SDK options and toggles.
+   * @param config.parseResponse If responses are parsed according to its `Content-Type` header.
+   */
+  config(config: ConfigOptions) {
+    this.core.setConfig(config);
+  }
+
+  /**
+   * If the API you're using requires authentication you can supply the required credentials
```
Dashron, line 26: It would be nice to have this comment auth aware, so it told you how to use it instead of a generic message (assuming I'm reading this file right as the output of the codegen)

packages/api/test/__fixtures__/sdk/petstore/index.ts
```diff
@@ -0,0 +1,617 @@
...
+   */
+  server(url: string, variables = {}) {
+    this.core.setServer(url, variables);
+  }
+
+  /**
+   * Add a new pet to the store
+   *
+   */
+  post<T = unknown>(path: string, body: Pet): Promise<T>;
```
Dashron, line 78: these all seem to be overloaded `post` methods (add, edit, pets, users, images, etc). how do we identify which one is actually being called?

packages/api/test/__fixtures__/sdk/petstore/index.ts
Dashron: at least it's codegen so they will be in sync. It would be nice if we could dedupe though with `@see` or something similar.

packages/api/test/__fixtures__/sdk/petstore/index.ts
```diff
@@ -0,0 +1,617 @@
...
+   */
+  loginUser(metadata: LoginUserMetadataParam): Promise<LoginUser_Response_200> {
+    return this.core.fetch('/user/login', 'get', metadata);
+  }
+
+  /**
+   * Get user by user name
+   *
+   */
+  getUserByName(metadata: GetUserByNameMetadataParam): Promise<User> {
```
Dashron, line 431: all these named methods are so nice and clean. I hope it will be one more DX reason to write better oas files.

PR #420, opened 2022-03-22
fix: cleanup of assorted snippet generation quirks

erunion:
## 🧰 Changes

This cleans up a handful of issues and quirks with the snippets that we're generating within `httpsnippet-client-api`:

* [x] 🧹 Slightly refactored how we were extracting path parameters out of URLs for snippets to instead re-use existing logic we have for this within `oas` and the `Oas.findOperation` method call we were already making. This has allowed me to not only drop the dependency `httpsnippet-client-api` had on `path-to-regexp` but also rely on the `oas` library to continue to maintain all of this logic instead of it being duplicated here too.
* [x] 🧹 Completely refactored how `httpsnippet-client-api` unit tests were structured so now it should be a lot easier to get up and go adding new tests.
* [x] 🧹  Moved all test fixtures over to referencing https://httpbin.org instead of http://mockbin.com as that's what we've generally been using for testing for a while now as it makes debugging calls **heaps** easier.
* [x] 🆕  Added a handful of new tests for auth handling that didn't exist previously. With the exception of cookie handling the snippet generator doesn't yet support `httpsnippet-client-api` now has 100% code coverage.
* [x] ✍️  Added some comments here and there within the library on pieces of code that I forgot what they were doing after not looking at this thing for well over a year. Hopefully they make sense.
* [x] 🐛  Fixed an issue where `sdk.auth()` lines in snippets didn't always end in a semicolon.
* [x] 🪣  Bumped `oas` to the latest release to pull in a few fixes:
  * [v18.0.4](https://github.com/readmeio/oas/releases/tag/18.0.4): Fixed an issue with how we were doing server variable matching on server variables that might be a hostname + subdomain.
  * [v18.0.5](https://github.com/readmeio/oas/releases/tag/18.0.5): `Operation.hasOperationId()` will now return `false` if the `operationId` is an empty string.
  * [v18.0.6](https://github.com/readmeio/oas/releases/tag/18.0.6): The new `camelCase` option in `Operation.getOperationId()` will now clean up an `operationId` if it can't be used as a method accessor.


packages/api/src/cache.ts
```diff
@@ -198,7 +198,7 @@ export default class Cache {
     return fetch(url)
       .then(res => {
         if (!res.ok) {
-          throw new Error(`Unable to retrieve URL. Reason: ${res.statusText}`);
+          throw new Error(`Unable to retrieve URL (${url}). Reason: ${res.statusText}`);
```
erunion, line 201: I had Mocha improperly set up in `httpsnippet-client-api` so that when I was running tests it was executing the snippet fixtures and dumping this error message into the console. Because it was extremely confusing to have "Unable to retrieve URL. Reason: 404" in the console adding the URL that it's failing to access here should help folks out when they encounter this error in the wild.

I also fixed the Mocha config to ignore those fixtures. 😳

packages/httpsnippet-client-api/src/index.js
```diff
@@ -53,7 +52,12 @@ function getAuthSources(operation) {
         if (scheme.in === 'query') {
           matchers.query.push(scheme.name);
         } else if (scheme.in === 'header') {
-          matchers.header[scheme.name] = '*';
+          // The way that this astrisk header matcher works is that since this `apiKey` goes in a
+          // named header (`scheme.name`) because the header is the key, we're matching against the
+          // entire header -- counter to the way that the HTTP basic matcher above works where we
+          // match and extract the API key from everything after `Basic ` in the `Authorization`
+          // header.
+          matchers.header[scheme.name.toLowerCase()] = '*';
```
erunion, line 60: Hope this comment makes sense because it took me a good bit staring this line and how it's used further below before understanding why I wrote it this way.

packages/httpsnippet-client-api/src/index.js
```diff
@@ -109,9 +98,10 @@ module.exports = function (source, options) {
   code.push(`const sdk = require('api')('${opts.apiDefinitionUri}');`);
   code.blank();
 
-  // If we have multiple servers configured and our source URL differs from the stock URL that we receive from our
-  // `oas` library then the URL either has server variables contained in it (that don't match the defaults), or the
-  // OAS offers alternate server URLs and we should expose that in the generated snippet.
+  // If we have multiple servers configured and our source URL differs from the stock URL that we
+  // receive from our `oas` library then the URL either has server variables contained in it (that
+  // don't match the defaults), or the OAS offers alternate server URLs and we should expose that
+  // in the generated snippet.
```
erunion, line 104: I've been trying to be a better small-screen citizen lately by capping my comments at the 100 character mark (traffic light yellow†) instead of going to a full 120. Hopefully you don't mind.

---
† traffic light yellow
```json
"[javascript]": {
  "editor.rulers": [
    80,
    { "column": 100, "color": "#F7B500" }, // traffic light yellow
    { "column": 120, "color": "#AA3F44" } // traffic light red
  ],
  "editor.tabSize": 2,
  "editor.defaultFormatter": "esbenp.prettier-vscode"
},
```

packages/httpsnippet-client-api/src/index.js
```diff
@@ -139,32 +129,27 @@ module.exports = function (source, options) {
     metadata = Object.assign(metadata, queryParams);
   }
 
-  // If we have path parameters present, we should only add them in if we have an operationId as we don't want metadata
-  // to duplicate what we'll be setting the path in the snippet to.
-  if ('operationId' in operation.schema) {
-    const pathParams = getParamsInPath(operation, operation.path);
```
erunion, line 145: Honestly really not sure why I duplicated the `getParamsInPath` work here from `oas` but `Oas.findOperation()` gives us everything we need for this work so 🤷

packages/httpsnippet-client-api/src/index.js
```diff
@@ -242,28 +229,20 @@ module.exports = function (source, options) {
   const args = [];
 
   let accessor = method;
-  if ('operationId' in operation.schema && operation.schema.operationId.length > 0) {
-    accessor = operation.schema.operationId;
+  if (operation.hasOperationId()) {
+    accessor = operation.getOperationId({ camelCase: true });
```
erunion, line 233: This `camelCase` option being here means that because this operation has an `operationId` we're going to do our best to try to clean it up so that it can be used as an accessor. I'm not going to recover the full logic it's doing but you can see it [here](https://github.com/readmeio/oas/blob/main/src/operation.ts#L326-L368).

What this work will allow us however to do is completely rid ourselves of the nasty array syntax that we had in place a bit below for non-alphanumeric IDs like the following:

```js
const sdk = require('api')('https://example.com/openapi.json');

sdk.auth('123');
sdk['find/pets-by-status']({status: 'available'})
```

That same snippet is now much cleaner:

```js
const sdk = require('api')('https://example.com/openapi.json');

sdk.auth('123');
sdk.findPetsByStatus({status: 'available'})
```

I will write tests in a followup PR within `api` to make sure that we're still able to handle backwards compatibility with the array'd operationID format though.

packages/httpsnippet-client-api/src/index.js
```diff
@@ -282,7 +261,7 @@ module.exports = function (source, options) {
   }
 
   code
-    .push(`sdk${accessor}(${args.join(', ')})`)
+    .push(`sdk.${accessor}(${args.join(', ')})`)
```
erunion, line 264: Moved the period out of `accessor` into here because it's cleaner that way and also without the array stuff that was on line 260 we can.

packages/httpsnippet-client-api/test/__datasets__/alternate-server/har.json
```diff
@@ -5,12 +5,6 @@
         "request": {
           "method": "POST",
           "url": "http://dev.local/v2/create/1234",
-          "headers": [
```
erunion, line 8: Removed this from this fixture because the [server variables spec](https://github.com/readmeio/oas-examples/blob/main/3.0/json/server-variables.json) in `@readme/oas-examples` doesn't have any auth and we have dedicated tests for asserting auth anyways.

packages/httpsnippet-client-api/test/__datasets__/issue-119/output.js
```diff
@@ -1,6 +1,6 @@
 const sdk = require('api')('https://example.com/openapi.json');
 
 sdk.auth('123');
-sdk['find/pets-by-status']({status: 'available'})
+sdk.findPetsByStatus({status: 'available'})
```
erunion, line 4: **Hella** cleaner.

packages/httpsnippet-client-api/test/__datasets__/operationid-with-underscores/output.js
```diff
@@ -0,0 +1,5 @@
+const sdk = require('api')('https://example.com/openapi.json');
+
+sdk.anythingOperation()
```
erunion, line 3: Though the `operationId` was `anything_Operation` we're cleaning it up to `anythingOperation` in `oas` because it's not alphanumeric ([code here](https://github.com/readmeio/oas/blob/main/src/operation.ts#L339)). Maybe underscores in these should be left alone?

packages/httpsnippet-client-api/test/index.test.js
```diff
@@ -6,6 +8,37 @@ const path = require('path');
 const client = require('../src');
 const readme = require('@readme/oas-examples/3.0/json/readme.json');
 
+const DATASETS_DIR = path.join(__dirname, '__datasets__');
+const SNIPPETS = require('fs').readdirSync(DATASETS_DIR);
+
+async function getSnippetDataset(snippet) {
```
erunion, line 14: Basic structure of a dataset test is the following:

```js
test/
  __datasets__/
    test-case-name/
      har.{js,json} // The HAR that we're generating a snippet for.
      openapi.{js,json} // The OpenAPI spec that we'll use to extract data for the URL in the HAR.
      output.js // The generated code snapshot.
```

packages/httpsnippet-client-api/test/index.test.js
```diff
@@ -75,81 +108,16 @@ describe('httpsnippet-client-api', function () {
     }).to.throw(/unable to locate a matching operation/i);
   });
 
-  describe('auth handling', function () {
```
erunion, line 78: Moved these auth handling tests into the `__datasets__` directory instead.

packages/httpsnippet-client-api/test/index.test.js
```diff
@@ -75,81 +108,16 @@ describe('httpsnippet-client-api', function () {
...
-        );
-
-        const definition = JSON.parse(
-          await fs.readFile(path.join(__dirname, `./__fixtures__/request/${request}/definition.json`), 'utf8')
-        );
-
-        const expected = await fs.readFile(path.join(__dirname, `./__fixtures__/output/${request}.js`), 'utf8');
+    SNIPPETS.forEach(snippet => {
+      it(`should generate \`${snippet}\` snippet`, async function () {
+        // Cookies test needs to get built out.
```
erunion, line 114: `api` supports cookies I just need to add support for it to our snippets.

packages/httpsnippet-client-api/src/index.js
Dashron: thoughts on indenting continued lines? I've always found it hard to know where one thought ends and another starts without it.


e.g.

```
  // If we have multiple servers configured and our source URL differs from the stock URL that we
  //     receive from our `oas` library then the URL either has server variables contained in it (that
  //     don't match the defaults), or the OAS offers alternate server URLs and we should expose that
  //     in the generated snippet.
```

Dashron: also would this be better as a /* */ block? do we have an eslint saying when to use one or the other?

Dashron (approved):
I think I looked at all the files that needed looking at? there's a lot of generated files here that didn't seem to need review.

packages/httpsnippet-client-api/src/index.js
erunion: I'm 1000% against indenting comments like this. I've also never personally liked using the `/* */` style of comment on comments that aren't a traditional method or class docblock. There's the [multiline-comment-style](https://eslint.org/docs/rules/multiline-comment-style) ESLint rule but it have any configuration for method/class docblock vs regular comments.

packages/httpsnippet-client-api/README.md
```diff
@@ -1,6 +1,6 @@
 # httpsnippet-client-api
 
-An HTTP Snippet client for generating snippets for the [api](https://npm.im/api) module.
+An HTTPSnippet client for generating snippets for the [api](https://npm.im/api) module.
```
kanadgupta, line 3: Wanna link to the `httpsnippet` package here?

packages/httpsnippet-client-api/src/index.js
```diff
@@ -53,7 +52,12 @@ function getAuthSources(operation) {
         if (scheme.in === 'query') {
           matchers.query.push(scheme.name);
         } else if (scheme.in === 'header') {
-          matchers.header[scheme.name] = '*';
+          // The way that this astrisk header matcher works is that since this `apiKey` goes in a
```
kanadgupta, line 55:
```suggestion
          // The way that this asterisk header matcher works is that since this `apiKey` goes in a
```

kanadgupta (approved):
Few small docs suggestions but overall LGTM! Thanks for the nice test coverage, I like how you organized it 🎉

PR #421, opened 2022-03-22
feat: adding support for cookies in snippets

erunion:
| 🚥 Fix RM-3946 |
| :-- |

## 🧰 Changes

This adds support for both cookie auth and cookie parameters in `api` snippets. This work was previously added to `api` in #393 but, contrary to what that PR says, snippets have never supported it.

## 🧬 QA & Testing

* [ ] See the test I added for cookie auth.
* [ ] I uncommented and fixed the existing cookie parameter test we had and that's passing now.

### DIFF
--- a/packages/httpsnippet-client-api/src/index.js
+++ b/packages/httpsnippet-client-api/src/index.js
@@ -115,19 +115,31 @@ module.exports = function (source, options) {
   }
 
   let metadata = {};
-  if (Object.keys(source.queryObj).length) {
-    const queryParams = source.queryObj;
+  Object.keys(source.queryObj).forEach(param => {
+    if (authSources.query.includes(param)) {
+      authData.push(buildAuthSnippet(source.queryObj[param]));
 
-    Object.keys(queryParams).forEach(param => {
-      if (authSources.query.includes(param)) {
-        authData.push(buildAuthSnippet(queryParams[param]));
+      // If this query param is part of an auth source then we don't want it doubled up in the
+      // snippet.
+      return;
+    }
 
-        delete queryParams[param];
-      }
-    });
+    metadata[param] = source.queryObj[param];
+  });
 
-    metadata = Object.assign(metadata, queryParams);
-  }
+  Object.keys(source.cookiesObj).forEach(cookie => {
+    if (authSources.cookie.includes(cookie)) {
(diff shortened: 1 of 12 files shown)


packages/httpsnippet-client-api/src/index.js
```diff
@@ -116,17 +116,33 @@ module.exports = function (source, options) {
...
   if (Object.keys(source.queryObj).length) {
-    const queryParams = source.queryObj;
-
-    Object.keys(queryParams).forEach(param => {
+    Object.keys(source.queryObj).forEach(param => {
       if (authSources.query.includes(param)) {
-        authData.push(buildAuthSnippet(queryParams[param]));
+        authData.push(buildAuthSnippet(source.queryObj[param]));
 
-        delete queryParams[param];
```
erunion, line 125: Not sure why I was doing this hokey property deletion stuff instead of just setting directly to `metadata`... 🤷

packages/httpsnippet-client-api/test/__datasets__/full-many-query-params/har.json
```diff
@@ -32,11 +32,11 @@
           ],
           "cookies": [
             {
-              "name": "foo",
+              "name": "foo-cookie",
```
erunion, line 35: Changed these cookie names off `foo` and `bar` because this, and the `full`, test have those already as parameters and `api` doesn't support differing parameter types that share the same name.

packages/httpsnippet-client-api/src/index.js
```diff
@@ -116,17 +116,33 @@ module.exports = function (source, options) {
...
+        // If this query param is part of an auth source then we don't want it doubled up in the
+        // snippet.
+        return;
       }
+
+      metadata[param] = source.queryObj[param];
     });
+  }
+
+  if (Object.keys(source.cookiesObj).length) {
```
kellyjosephprice, line 132: What is this `length` check preventing? Wouldn't the `forEach` essentially be a no-op on an empty array?

Dashron (approved):
code and tests look fine

kellyjosephprice (approved):
LGTM.

packages/httpsnippet-client-api/src/index.js
erunion: Great point.

PR #422, opened 2022-03-23
feat: creating a new test suite for running snippet datasets within a VM

erunion:
| 🚥 Fix RM-3945, RM-3986 |
| :-- |

## 🧙  Changes

This introduces the creation of a new addition to the `httpsnippet-client-api` test suite to ensure that the `api` snippets that we're generating work how we expect they to work, and don't just *look* okay.

Because the way that these snippet fixtures/datasets are stored as JS files within the `test/__datasets__` directory it introduced an interesting challenge of how to execute JS code, with contextual [nock](https://npm.im/nock) interceptors, and an assertion that it returns an expected response.

Introducing [Node VM's](https://nodejs.org/api/vm.html).

Before landing on this solution I attempted a couple things:

* Dynamically wrapping these snippet output datasets in a `module.exports = () => { // sdk code }` call. This just flat out didn't work at all. The nock interceptors that I was setting up outside of the `require()` wouldn't be called, and neither would our `console.log` stubs be recognized.
* Running the code through `eval`. This time the nocks were being recognized but our `console.log` stub wasn't being used.

There's a bit of work involved to set up a Node VM within the test, but it allows me to do everything I wasn't able to do with the above solutions:

* Create interceptor datasets for each snippet, feed them into `nock` and have them used.
* Capture output from `console.log()` within the snippet, inside the VM, so I can assert that not only was the HTTP request called form the nock but that I got back what I was expecting.

I can't that believe this all works either.

## 📚  How do I write one of these snippet tests?

I've created a new `CONTRIBUTING` guide in the repo with details on how to do this.

## 👩‍🔧  Additional fixes

In addition to that madness I also fixed a couple bugs in `api` that surfaced themselves while writing this test suite:

* [x] If a request body payload consisted of only a nullish property (eg. `{ foo: null }`) when it would get processed through [@readme/oas-to-har](https://npm.im/@readme/oas-to-har) `foo` would be deleted out of the payload resulting in a corrupted HAR with an `undefined` `postData.text` entry -- this would then cause the [fetch-har](https://npm.im/fetch-har) to fatally terminate.
  * See https://github.com/readmeio/remove-undefined-objects/pull/8 and https://github.com/readmeio/oas-to-har/pull/70 for these fixes.
* [x] Under some circumstances if we ran `Operation.prepareSecurity()` in the [oas](https://npm.im/oas) library and then ran spec validation with [@readme/openapi-parser](https://npm.im/@readme/openapi-parser) validation would fail because a part of the `prepareSecurity()` library was mistakenly updating the internal OAS variable reference with internal data.
  * See https://github.com/readmeio/oas/pull/627 for this fix.
* [x] Though `api` received support for cookie parameters in #393  and #421 I completely forgot to add support for it to auth handling.
* [x] The new operation IDs that we're now creating with `Operation.getOperationId({ camelCase: true })` weren't being recognized by `api` and would throw a fatal error saying the operation couldn't be found.
* [x] A handful of tests had data in their HAR that didn't map to anything within the OpenAPI file it was attached to.

## 🧬 QA & Testing

If the tests in this suite aren't passing then I've got egg on my face. 🙃


packages/httpsnippet-client-api/test/index.test.js
```diff
@@ -1,42 +1,64 @@
...
 const DATASETS_DIR = path.join(__dirname, '__datasets__');
 const SNIPPETS = require('fs').readdirSync(DATASETS_DIR);
 
 async function getSnippetDataset(snippet) {
+  const SNIPPET_DIR = path.join(DATASETS_DIR, snippet);
+
   const har = await fs
-    .stat(path.join(DATASETS_DIR, snippet, 'har.json'))
-    .then(() => require(path.join(DATASETS_DIR, snippet, 'har.json')))
+    .stat(path.join(SNIPPET_DIR, 'har.json'))
```
erunion, line 26: There's a lot of boilerplate in this method, if you have ideas for how to clean it up I'm all 👂

rafegoldberg (approved):
So this is definitely above-my-paygrade level stuff. But from what I understand of what's going down here the code looks solid!

packages/httpsnippet-client-api/test/index.test.js
ilias-t: assigning `path.join(SNIPPET_DIR, 'har.json')` to a variable and reusing that a bit might help

packages/httpsnippet-client-api/test/index.test.js
```diff
@@ -1,42 +1,64 @@
 /* eslint-disable global-require */
 /* eslint-disable import/no-dynamic-require */
 /* eslint-disable mocha/no-setup-in-describe */
+const chai = require('chai');
 const { expect } = require('chai');
+const sinon = require('sinon');
+const sinonChai = require('sinon-chai');
+const vm = require('vm');
```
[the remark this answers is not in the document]
erunion: My **extremely** baseline understanding is that `vm` creates a new instance of the V8 engine and runs whatever you pass into it, with the context (read: variables) you set up.

Yes I'm scared to breathe on all of this for fear that it'll fall apart.

packages/api/src/core/prepareAuth.ts
```diff
@@ -72,7 +72,7 @@ export default function prepareAuth(authKeys: (number | string)[][], operation:
           );
         }
 
-        if (scheme.in === 'query' || scheme.in === 'header') {
+        if (scheme.in === 'query' || scheme.in === 'header' || scheme.in === 'cookie') {
```
erunion, line 75: This missing line was preventing cookie-based usages of `sdk.auth()` from working.

packages/api/src/index.ts
```diff
@@ -59,14 +59,16 @@ class Sdk {
...
-          // to know here if we actually have one present or not.
-          return operation.hasOperationId() ? operation.getOperationId() : false;
-        })
+        .filter(operation => operation.hasOperationId())
         .reduce((prev, next) => {
+          // `getOperationId()` creates dynamic operation IDs when one isn't available but we need
+          // to know here if we actually have one present or not. The `camelCase` option here also
+          // cleans up any `operationId` that we might have into something that can be used as a
+          // valid JS method.
+          const operationId = next.getOperationId({ camelCase: true });
```
erunion, line 68: Not using the `getOperationId()` `camelCase` option anywhere in this block was causing any operation that that method was reshaping into a usage JS accessor as unrecognized.

packages/api/test/index.test.ts
```diff
@@ -67,9 +67,9 @@ describe('api', function () {
         'head',
         'patch',
         'trace',
-        'list-data-sets',
-        'list-searchable-fields',
-        'perform-search',
+        'listDataSets',
```
erunion, line 70: This is part of the work that the `camelCase` option in `getOperationId()` does by making something that can't be used as a JS accessor without arrayifying it (eg. `sdk['list-data-sets']()`) into the **much** cleaner `sdk.listDataSets()`.

packages/httpsnippet-client-api/test/__datasets__/auth-query/har.json
```diff
@@ -7,10 +7,6 @@
           "httpVersion": "HTTP/1.1",
           "method": "GET",
           "queryString": [
-            {
```
erunion, line 10: Removed this because the OpenAPI definition that this `auth-query` test uses doesn't have a `status` query parameter defined.

packages/httpsnippet-client-api/test/__datasets__/full-many-query-params/openapi.json
```diff
@@ -36,6 +36,20 @@
               "type": "string"
             },
             "required": true
+          },
+          {
```
erunion, line 40: Added these because this `full-many-query-params` test is testing some cookie parameters but because they weren't defined here as a cookie parameter they were instead getting sent as query parameters.

packages/httpsnippet-client-api/test/__datasets__/issue-128/har.json
```diff
@@ -3,31 +3,17 @@
...
-          "headers": [
-            {
-              "name": "Accept",
-              "value": "application/xml"
-            }
-          ],
+          "headers": [],
           "httpVersion": "HTTP/1.1",
           "method": "GET",
-          "postData": {
```
erunion, line 18: This test is testing a GET request so it doesn't have any way of having a POST payload. Also removed the `Accept` headers from this because the OpenAPI definition for this one doesn't have anything about XML in it.

packages/httpsnippet-client-api/test/__datasets__/issue-76/har.json
```diff
@@ -3,25 +3,9 @@
     "entries": [
       {
         "request": {
-          "allHeaders": {
-            "accept": "application/xml",
-            "authorization": "Bearer 123"
```
erunion, line 8: Removed this because the auth for this test is defined as a query string. `allHeaders` is also specific to [httpsnippet](http://npm.im/httpsnippet) so we weren't even using this in the snippet anyways.

packages/httpsnippet-client-api/test/__datasets__/multipart-data/har.json
```diff
@@ -17,7 +17,7 @@
               {
                 "name": "foo",
                 "value": "Hello World",
-                "fileName": "hello.txt",
+                "fileName": "test/__fixtures__/files/hello.txt",
```
erunion, line 20: Changed this file path so that the `hello.txt` fixture I added in this PR will get picked up and sent with the snippet.

packages/httpsnippet-client-api/test/__datasets__/multipart-data/mock.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  scope: 'https://httpbin.org',
+  method: 'POST',
+  path: '/anything',
+  body: /--form-data-boundary-(.*)\r\nContent-Disposition: form-data; name="foo"; filename="hello.txt"\r\nContent-Type: text\/plain\r\n\r\nHello world!\n\r\n--form-data-boundary-(.*)--\r\n\r\n/,
```
erunion, line 5: `multipart/form-data` boundaries are always unique so I needed to regex this.

The contents of this payload are coming from the contents of `test/__fixtures__/files/hello.txt` as well. File uploads, they work!

packages/httpsnippet-client-api/test/index.test.js
```diff
@@ -1,42 +1,64 @@
 /* eslint-disable global-require */
 /* eslint-disable import/no-dynamic-require */
 /* eslint-disable mocha/no-setup-in-describe */
+const chai = require('chai');
```
erunion, line 4: If you're curious why I'm using [chai](https://npm.im/chai) and [mocha](https://mochajs.org/) within this repo instead of Jest it's because I've been slowly moving our open source projects that have code that can be run in a browser to have tests using Chai + Mocha as you can run the same tests in a real browser with [Karma](https://karma-runner.github.io/latest/index.html) with **extremely** minimal effort.

Projects that have been migrated over (so far) are [fetch-har](https://github.com/readmeio/fetch-har) and [openapi-parser](https://github.com/readmeio/openapi-parser).

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,208 @@
+# Contributing
```
erunion, line 1: @kanadgupta You should review this contributing guide I'm adding to the repo.

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,208 @@
...
+## Commit Conventions
+
+For our general commit conventions please consult our organization contributing guidelines [here](https://github.com/readmeio/.github/blob/main/.github/CONTRIBUTING.md).
+
+## Help
+
+If you need help don't be afraid to open up an issue!
+
+## Packages
+### `httpsnippet-client-api` snippet
```
kanadgupta, line 14: Wanna run this file through your VS Code formatter? I think there should be a newline between these headers

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,208 @@
...
+
+If you wish to add or change functionality to how `api` snippets are generated, your changes should go within the `httpsnippet-client-api` package in `packages/`. We have a nifty test suite set up there that'll run pre-made JS code snippets for the [api](https://npm.im/api) package within a virtual environment, with mocked out HTTP requests, to ensure that everything is well and good. It's pretty neat!
+
+To add a new test into this suite you should do the following:
+
+1. Create a new directory for your test in `packages/httpsnippet-client-api/test/__datasets__`. This directory should be a short and clear description of what you're testing.
+    * If you need to give additional context to your test please do add a `README` in this directory!
```
kanadgupta, line 21:
```suggestion
    * If you need to give additional context to your test, please add a `README` in this directory!
```

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,208 @@
...
+2. In this new directory you'll need to create 4 files: `har.json`, `openapi.json`, `mock.json`, and `output.js`.
+
+#### har.json
```
kanadgupta, line 24:
```suggestion
#### `har.json`
```

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,208 @@
...
+
+The way that `httpsnippet-client-api` works is that it takes a HAR object and constructs a snippet according to it, and what the matching operation in an OpenAPI definition, say the API request should function.
+
+This file should contain a fully defined [HAR object](http://www.softwareishard.com/blog/har-12-spec/) for your API request.
+
+If you need to use dynamic data (like base64 encoding an API key) then this file can be renamed to `har.js` and the HAR can be exported with `module.exports`.
+
+#### openapi.json
```
kanadgupta, line 32:
```suggestion
#### `openapi.json`
```

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,208 @@
...
+
+If you'd like to reference one of the definitions that we provide in the [@readme/oas-examples](https://npm.im/@readme/oas-examples) package then you can rename `openapi.json` to `openapi.js` and have a single line of the following and the test suite will pick it up!
+
+```js
+module.exports = require('@readme/oas-examples/path/to/spec');
+````
+
+Additionally if you need help verifying that your OpenAPI definition is valid you can run `npx rdme validate path/to/your/definition.json`.
+
+#### mock.json
```
kanadgupta, line 44:
```suggestion
#### `mock.json`
```

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,208 @@
...
+
+Some things you should be aware of:
+
+* You can leave `response` out of `mock.json` as our test suite automatically mocks out a `response` to use so we can assert that what we mocked as a response was what we got.
+* If your API doesn't have a request body you can leave `body` off.
+* If your API request doesn't use any unique headers (eg. `Authentication` or `Content-Type`), then you can generally leave `reqheaders` off.
+* `status` should always be 200.
+* If you're testing something like `multpart/form-data` requests and need to regex a boundary you can rename `mock.json` to `mock.js` and add your regex in. See the `multipart-*` test for examples of this.
+
+#### output.js
```
kanadgupta, line 201:
```suggestion
#### `output.js`
```

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,208 @@
...
+
+To add a new test into this suite you should do the following:
+
+1. Create a new directory for your test in `packages/httpsnippet-client-api/test/__datasets__`. This directory should be a short and clear description of what you're testing.
+    * If you need to give additional context to your test please do add a `README` in this directory!
+2. In this new directory you'll need to create 4 files: `har.json`, `openapi.json`, `mock.json`, and `output.js`.
+
+#### har.json
+
+The way that `httpsnippet-client-api` works is that it takes a HAR object and constructs a snippet according to it, and what the matching operation in an OpenAPI definition, say the API request should function.
```
kanadgupta, line 26: This reads a bit weirdly so I tried rewriting it, but feel free to mess with it further:
```suggestion
`httpsnippet-client-api` works by taking a HAR object and its matching operation in an OpenAPI definition and constructing a code snippet based upon them.
```

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,208 @@
...
+
+This file should contain a fully defined [HAR object](http://www.softwareishard.com/blog/har-12-spec/) for your API request.
+
+If you need to use dynamic data (like base64 encoding an API key) then this file can be renamed to `har.js` and the HAR can be exported with `module.exports`.
+
+#### openapi.json
+
+Within `openapi.json` you'll want to spec out the OpenAPI definition that will be powering this request. This definition should be fully valid to the spec.
```
kanadgupta, line 34: Idea I just had—since we require that every spec is valid, would it make sense to have a generic `beforeEach` test that asserts this using our validator?

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,208 @@
...
+
+If you'd like to reference one of the definitions that we provide in the [@readme/oas-examples](https://npm.im/@readme/oas-examples) package then you can rename `openapi.json` to `openapi.js` and have a single line of the following and the test suite will pick it up!
```
kanadgupta, line 36: I noticed that you did this and this is really nice! What are your thoughts on us doing this kind of thing in our main codebase? I noticed a PR the other day where you were manually incorporating updates from the `oas-examples` repo and this pattern might help us avoid that?

kanadgupta:
```suggestion
If you'd like to reference one of the definitions that we provide in the [`@readme/oas-examples`](https://npm.im/@readme/oas-examples) package then you can rename `openapi.json` to `openapi.js` and have a single line of the following and the test suite will pick it up!
```

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,208 @@
...
+
+```js
+module.exports = require('@readme/oas-examples/path/to/spec');
+````
+
+Additionally if you need help verifying that your OpenAPI definition is valid you can run `npx rdme validate path/to/your/definition.json`.
```
kanadgupta, line 42: 😎

kanadgupta: this also kinda ties into my comment above about validating specs in our tests

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,208 @@
...
+
+#### mock.json
+
+This file is what is going to tell our test suite what endpoints your API test will be accessing so we can mock out those requests for you. Behind the scenes we use the amazing [nock](https://npm.im/nock) library for this. Generating this data is a bit difficult, but you can use `nock` itself to guide you along.
```
kanadgupta, line 46:
```suggestion
This file is what is going to tell our test suite what endpoints your API test will be accessing so we can mock out those requests for you. Behind the scenes we use the amazing [`nock`](https://npm.im/nock) library for this. Generating this data is a bit difficult, but you can use `nock` itself to guide you along.
```

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,208 @@
...
+  args: { baz: 'abc', foo: [ 'bar', 'baz' ], key: 'value' },
+  data: '',
+  files: {},
+  form: {},
+  headers: {
+    Accept: '*/*',
+    'Accept-Encoding': 'gzip,deflate',
+    Host: 'httpbin.org',
+    'User-Agent': 'api (node)/4.2.0',
+    'X-Amzn-Trace-Id': 'REDACTED'
```
kanadgupta, line 71: weird lol, is this an httpbin thing?

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,208 @@
...
+}
+```
+
+To have `nock` interpret this request into data that we can use to mock out requests back with `nock` you can add modify your snippet as such:
+
+```js
+const nock = require('nock');
+
+nock.recorder.rec({
+  output_objects: true,
```
kanadgupta, line 86: Interesting, I've never used this `output_objects` option before! I really dig this approach, makes writing tests a loooot quicker

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,208 @@
...
+  "path": "/anything?key=value&foo=bar&foo=baz&baz=abc",
+  "status": 200
+}
+```
+
+Some things you should be aware of:
+
+* You can leave `response` out of `mock.json` as our test suite automatically mocks out a `response` to use so we can assert that what we mocked as a response was what we got.
+* If your API doesn't have a request body you can leave `body` off.
+* If your API request doesn't use any unique headers (eg. `Authentication` or `Content-Type`), then you can generally leave `reqheaders` off.
```
kanadgupta, line 197:
```suggestion
* If your API request doesn't use any unique headers (e.g. `Authentication` or `Content-Type`), then you can generally leave `reqheaders` off.
```

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,208 @@
...
+* If you're testing something like `multpart/form-data` requests and need to regex a boundary you can rename `mock.json` to `mock.js` and add your regex in. See the `multipart-*` test for examples of this.
+
+#### output.js
+
+Now that you've done all the hard work for setting up your unit test all that's left is to craft the snippet you're expecting we should generate for your use case. You have a couple options for this:
+
+1. Write it manually.
+2. Run the test suite without one so it fails, and then populate what we thought it should be from the data in `har.json` and `openapi.json` into `output.js`.
+
+After that, and all your tests are passing, then you're good! Thanks for helping out!
```
kanadgupta, line 208: This is super thorough and I learned a ton by just reading this—thanks for doing this!

packages/api/test/core/prepareAuth.test.ts
```diff
@@ -49,6 +49,26 @@ describe('#prepareAuth()', function () {
...
+        });
+      });
+
+      it('should throw if you supply multiple auth keys', function () {
+        const operation = oas.operation('/apiKey', 'post');
+        const authKeys = [[apiKey, apiKey]];
+
+        expect(() => {
+          prepareAuth(authKeys, operation);
+        }).to.throw(/only a single key is needed/i);
```
kanadgupta, line 69: How difficult would it be to do an exact match here?

packages/httpsnippet-client-api/test/index.test.js
kanadgupta: people need to chill out with their pseudo-ethnic package naming conventions lmao

packages/httpsnippet-client-api/test/index.test.js
```diff
@@ -110,16 +132,106 @@ describe('httpsnippet-client-api', function () {
...
+          consoleStub.restore();
         });
 
-        expect(`${code}\n`).to.equal(expected);
+        it('should generate the expected snippet', async function () {
+          const [har, definition] = await getSnippetDataset(snippet);
+          const expected = await fs.readFile(path.join(DATASETS_DIR, snippet, 'output.js'), 'utf-8');
+
+          const code = new HTTPSnippet(har).convert('node', 'api', {
+            apiDefinitionUri: 'https://example.com/openapi.json',
```
kanadgupta, line 160: Why is the path here `/openapi.json` and not `/openapi-${snippet}.json`?

kanadgupta (changes requested):
Sorry for all the comments, most of them are nits. I'm obsessed with the fact that this test suite also caught a ton of bugs in the process 😻 amazing work here!

.github/CONTRIBUTING.md
erunion: I definitely want to do that but I need to figure out a way to incorporate that with the `packages/testing-utils` package compilation process and how https://preview.readme.io uses that.

.github/CONTRIBUTING.md
erunion: Yeah httpbin is served from AWS I guess and adds this into the headers that come back in its responses. Probably from a gateway sitting in front of their servers?

packages/httpsnippet-client-api/test/index.test.js
erunion: No particular reason but I'll update this so it's a bit more contextual to each snippet and more in line with how the snippet VM runner mocks them out.

kanadgupta: Gotcha! I'm only scanning this test but I'm surprised it passes, considering [the nock is different](https://github.com/readmeio/api/pull/422/files/bcbfac97bcc8aeef2a1e07d9c65ef4097a095b11#diff-ae672312e784c6f29f666bc87496973fde59885728afef764ac01a128cc75b56R173)

erunion: I'm updating them to `https://api.example.com/{snippet}.json`

.github/CONTRIBUTING.md
erunion: Added pre-test validation!

kanadgupta (approved):
LGTM! 🚀

Issue #102, opened 2020-07-06
`prepareAuth` errors with multiple requests


kanadgupta:
I'm seeing an error when making multiple authenticated requests as shown below:
```js
const sdk = require('api')('https://docs.readme.com/developers/openapi/5e1a4f6284248b0018149174');

async function run() {
  const changelogs = await sdk.auth('API_KEY').getChangelogs()
    .then(res => res.json())
    .then(res => {
      return res;
    });

  return sdk.auth('API_KEY').getChangelog({ slug: changelogs[0].slug })
    .then(res => res.json())
    .then(res => {
      console.log(res);
    });
}

run();
```
I get this error:
```
TypeError: Cannot read property 'length' of undefined
    at authKeys.forEach (code/node_modules/api/src/lib/prepareAuth.js:21:17)
    at Array.forEach (<anonymous>)
    at module.exports (code/node_modules/api/src/lib/prepareAuth.js:19:12)
    at fetchOperation (code/node_modules/api/src/index.js:37:87)
    at Object.assign.args (code/node_modules/api/src/index.js:63:22)
    at Proxy.<anonymous> (code/node_modules/api/src/index.js:117:33)
    at run (code/index.js:10:55)
```
For some reason, it's having trouble loading `securitySchemes` here:
https://github.com/readmeio/api/blob/cc67a0cc8415beb86e0995f970161266c74fe4d5/packages/api/src/lib/prepareAuth.js#L21

domharrington:
Failing test for this here: https://github.com/readmeio/api/commit/56ce53e08d590d4a532a494e86e46a50efc4d891

I dont think it's possible for us to know the difference between a global call to auth, and a local call to auth:

```
sdk.auth(apiKey);

await sdk
  .auth(apiKey)
  .getSomething();
```

Since there's shared auth state, doing the following mutates the global state and doesn't work:

```
await sdk
  .auth(apiKey1)
  .getSomething()
  .then(() => mock1.done());

await sdk
  .auth(apiKey2)
  .getSomething()
  .then(() => mock2.done());
```

I think short of renaming the global/local functions, there's not much we can do here. We could possibly use a [Proxy.set()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/Proxy/set) and do something like this:

```
// global auth via proxy setter
sdk.auth = apiKey

// local auth via auth() method chaining
await sdk
  .auth(apiKey)
  .getSomething();
```

[...]

PR #431, opened 2022-04-11
fix(auth): remove the ability to have auth chained with each request

domharrington:
## 🧰 Changes
There's [a bug](https://github.com/readmeio/api/issues/102) with chaining auth calls due to the fact that we have a global state, and calling auth() again appends on to the state and breaks prepareAuth.

This PR removes the ability to call auth() in a chainable fashion, and only allows auth to be called at the top level.

I'd like to figure out a way to bring this back eventually, but the current API design doesn't support it in a clean way. We can revisit this when we look to support operations with multiple auth mechanisms.

This should probably be a major version bump due to removing functionality.

Fixes https://github.com/readmeio/api/issues/102

## 🧬 QA & Testing

Does CI pass? 🟢

### DIFF
--- a/packages/api/src/core/index.ts
+++ b/packages/api/src/core/index.ts
@@ -26,7 +26,7 @@ export { getJSONSchemaDefaults, parseResponse, prepareAuth, prepareParams, prepa
 export default class APICore {
   spec: Oas;
 
-  private auth: (number | string)[][] = [];
+  private auth: (number | string)[] = [];
 
   private server:
     | false
@@ -59,7 +59,7 @@ export default class APICore {
   }
 
   setAuth(...values: string[] | number[]) {
-    this.auth.push(values);
+    this.auth = values;
     return this;
   }
 
--- a/packages/api/src/core/prepareAuth.ts
+++ b/packages/api/src/core/prepareAuth.ts
@@ -3,8 +3,8 @@ import type { Operation } from 'oas';
 
 type SecurityType = 'Basic' | 'Bearer' | 'Query' | 'Header' | 'Cookie' | 'OAuth2' | 'http' | 'apiKey';
 
-export default function prepareAuth(authKeys: (number | string)[][], operation: Operation) {
-  if (authKeys.length === 0) {
+export default function prepareAuth(authKey: (number | string)[], operation: Operation) {
+  if (authKey.length === 0) {
     return {};
   }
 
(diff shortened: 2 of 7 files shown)


packages/api/src/core/index.ts
```diff
@@ -26,7 +26,7 @@ export { getJSONSchemaDefaults, parseResponse, prepareAuth, prepareParams, prepa
 export default class APICore {
   spec: Oas;
 
-  private auth: (number | string)[][] = [];
+  private auth: (number | string)[] = [];
```
domharrington, line 29: Did i typescript correctly?

packages/api/src/core/index.ts
```diff
@@ -59,7 +59,7 @@ export default class APICore {
   }
 
   setAuth(...values: string[] | number[]) {
-    this.auth.push(values);
+    this.auth = values;
     return this;
```
domharrington, line 63: Not sure why these are all returning `this`. I don't think it needs to, but I kept it to be consistent with the others.

packages/api/src/core/prepareAuth.ts
```diff
@@ -27,60 +27,59 @@ export default function prepareAuth(authKeys: (number | string)[][], operation:
...
-        break;
+  const securityType = securitySchemes[0] as SecurityType;
+
+  const schemes = security[securityType];
+
+  if (schemes.length > 1) {
+    throw new Error("Sorry, this API currently requires multiple forms of authentication which we don't yet support.");
+  }
+
+  const scheme = schemes[0];
```
domharrington, line 38: Same as https://github.com/readmeio/api/pull/431/files#r847494000=

packages/api/src/index.ts
```diff
@@ -156,7 +156,6 @@ class Sdk {
        */
       auth: (...values: string[] | number[]) => {
         core.setAuth(...values);
-        return new Proxy(sdk, sdkProxy);
```
domharrington, line 159: This is the part that actually removes the chaining.

domharrington (commented):
- [ ] Can I remove this part here as well? https://github.com/readmeio/api/blob/15bfc55bfecdff4d336fa632a2e6370008e77d3b/packages/api/src/index.ts#L104
- [ ] I've left the tests skipped, up to you if you wanna keep em around or if we should just remove them

packages/api/src/core/index.ts
erunion: Yep. I updated `prepareAuth.test.ts` in https://github.com/readmeio/api/pull/431/commits/091049816549923dc1069df2ea6d679fa4d075f1 to account for this.

packages/api/src/core/index.ts
erunion: I think this was my OOP brain trying to make things chainable and I don't think any of these `set*(` methods need to be.

packages/api/src/core/prepareAuth.ts
```diff
@@ -27,60 +27,59 @@ export default function prepareAuth(authKeys: (number | string)[][], operation:
...
-          if (authKey.length > 1) {
-            throw new Error(
-              'Multiple auth tokens were supplied for the auth on this endpoint, but only a single token is needed.'
-            );
-          }
-
-          preparedAuth[scheme._key] = authKey[0];
-        }
-        break;
+  const securityType = securitySchemes[0] as SecurityType;
```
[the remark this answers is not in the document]
erunion: Yeah this is fine.

packages/api/src/index.ts
erunion: Can you update the README? There's a handful of references in there talking about how `.auth()` is chainable. You'll need to update the root README and the one in `packages/api` too but I believe the root one is just a copy of the package one.

packages/api/test/auth.test.ts
```diff
@@ -29,30 +29,29 @@ describe('#auth()', function () {
...
+        const mock = nock('https://httpbin.org')
+          .get(/\/anything\/apiKey/)
+          .reply(200, uri => uri);
+
+        sdk.auth(apiKey);
+        expect(await sdk.get('/anything/apiKey')).to.equal('/anything/apiKey?apiKey=123457890');
+        mock.done();
+      });
+
+      it.skip('should allow you to supply auth when chained with an operation', async function () {
```
erunion, line 42: I'm good with removing all these skipped tests.

erunion (changes requested):
need to update the docs on `.auth()` but otherwise this looks good.

TomMiller-mas:
Late to the party, but what about letting us close the connection manually when we are done with the SDK connection?  

`await sdk.close();`

This allows us to leave it open as long as we want to.

PR #444, opened 2022-05-17
feat: configurable cache dir

erunion:
> Do not merge this PR in, I will tag a v4.5 directory off of this branch.

## 🧰 Changes

Because the `node_modules/.cache/api` directory system has become a growing problem in certain environments that prune `node_modules/` out I'm adding a new `cacheDir` option to the dynamic API SDK instantiation that will let folks configure where they want our caching system to save its data to.

Much of this work is an early precursor to how the new codegen SDK storage will work in v5.

## 🔖  Reviewing this

Because this PR is a bit of a mess it'll be easier if you review the only two commits that I made for this work:

* [ ] https://github.com/readmeio/api/pull/444/commits/d4a7df3bc80ebe44a8bc759ad3e209f244058513
* [ ] https://github.com/readmeio/api/pull/444/commits/88637911e7584892d68930499c0822ff142cbc1f

## 🧬 QA & Testing

You can test this locally by running this code from within `packages/api`:

```js
const path = require('path');
const sdk = require('./src')('https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/yaml/readme.yaml', {
  cacheDir: path.join(process.cwd(), '.api'),
});

sdk.getOpenRoles().then(res => {
  console.log(`there are ${res.length} open roles`);
  console.log(res[0]);
});
```

After running this you'll have a `.api/` directory in `packages/api` that will have a `cache.json` that looks like this:

Screen Shot 2022-05-17 at 1 57 19 PM

And head'ing the spec in `.api/specs/` will be the ReadMe spec that was pulled down from https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/yaml/readme.yaml:

Screen Shot 2022-05-17 at 1 58 15 PM

### DIFF
--- a/packages/api/README.md
+++ b/packages/api/README.md
@@ -23,7 +23,7 @@ npm install api --save
 Using `api` is as simple as supplying it an OpenAPI and using the SDK as you would any other!
 
 ```js
-const sdk = require('api')('https://raw.githubusercontent.com/readmeio/oas/master/packages/examples/3.0/json/petstore.json');
+const sdk = require('api')('https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/json/petstore.json');
 
 sdk.listPets().then(res => {
   console.log(`My pets name is ${res[0].name}!`);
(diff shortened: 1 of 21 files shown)


packages/api/README.md
```diff
@@ -161,3 +164,17 @@ By default we parse the response based on the `content-type` header for you. You
 ```js
 sdk.config({ parseResponse: false });
 ```
+
+#### Where is the cache stored?
```
erunion, line 168: This is the only change in this PR for this file.

PR #447, opened 2022-05-25
fix: use existing API methods

bennycode:
| 🚥 Fix RM-XXX |
| :-- |

## 🧰 Changes

There is no "operationId" with the name of "listPets" in https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/json/petstore.json, so when calling the API example you will run into an error:

```
Error: Sorry, `listPets` does not appear to be a valid operation on this API.
```

To fix that I am using "findPetsByStatus" which I have verified with: https://petstore.swagger.io/#/pet/findPetsByStatus

Unfortunately, the response from "findPetsByStatus" comes back as "application/xml". Do you know how I can explicitly request "application/json" using this "api" package?

## 🧬 QA & Testing

Just run this API example:

```ts
const sdk = require('api')('https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/json/petstore.json');
sdk.findPetsByStatus({ status: "available" }).then(async (res) => {
  console.log(`My pets name is ${res[0].name}!`);
}).catch(err => {
  console.error(err);
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ All you need to use `api` is to supply it an OpenAPI definition and then use the
 ```js
 const sdk = require('api')('https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/json/petstore.json');
 
-sdk.listPets().then(res => {
+sdk.findPetsByStatus({ status: 'available' }).then(res => {
   console.log(`My pets name is ${res[0].name}!`);
 });
 ```
@@ -37,7 +37,7 @@ The OpenAPI definition is automatically downloaded, cached, and transformed into
 
 ```js
 sdk.auth('myApiToken');
-sdk.listPets().then(...);
+sdk.findPetsByStatus({ status: 'available' }).then(...);
 ```
 
 With the exception of OpenID, it supports all forms of authentication supported by the OpenAPI specification! Supply `.auth()` with your auth credentials and it'll magically figure out how to use it according to the API you're using. 🧙‍♀️
@@ -173,7 +173,7 @@ By default the cache is configured with the [find-cache-dir](https://npm.im/find
 const sdk = require('api')('https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/json/petstore.json', {
   cacheDir: './path/to/my/custom/cache/dir',
 });
-sdk.listPets().then(res => {
+sdk.findPetsByStatus({ status: 'available' }).then(res => {
   console.log(`My pets name is ${res[0].name}!`);
 });
 ```


domharrington:
Hey! Thanks for spotting that missing `operationId`, i think `listPets` used to be present in an older version of the petstore spec and that's where that came from. 

As for the xml content type returned, that's a tricky edge case. We do have a 2nd parameter that you can pass to the api function call called metadata, which should make it possible to override some of these things using something like:

```js
const sdk = require('api')('https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/json/petstore.json');

sdk.findPetsByStatus({ status: 'available' }, { Accept: 'application/json' }).then(async (res) => {
  console.log(res);
}).catch(err => {
  console.error(err);
});
```

But because the spec file doesn't document that you can provide an Accept header, we appear to be just stripping that out: https://github.com/readmeio/api/blob/5c3490291498c1ac7f2f9dfe2ad99afff9010689/packages/api/src/lib/prepareParams.js#L162-L165

Then upstream we're setting the Accept header to match the first response content type:
https://github.com/readmeio/oas-to-har/blob/8e9cda68c37f575f2f7b0e541a9df2712f69a3c2/src/index.js#L260-L263

Which in this case they've documented application/xml as the first one: 
https://github.com/readmeio/oas-examples/blob/dbbd22c147f5f79ba87c355332da064f98373630/3.0/json/petstore.json#L121-L128

The api module is designed to abstract away a lot of these complexities for you by picking sensible defaults, but unfortunately in this case the escape hatch mechanism isn't sufficient to bypass those. I'll wait to hear what @erunion has to say on this but this is definitely a valid issue here. If it is in fact possible to do this via the metadata object that we already have, then we need much better documentation around that.

Once again, thanks for raising.

domharrington:
Another issue appears to be that as soon as you add the 2nd metadata argument like so:

```js
const sdk = require('api')('https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/json/petstore.json');

sdk.findPetsByStatus({ status: 'available' }, {}).then(async (res) => {
  console.log(res);
}).catch(err => {
  console.error(err);
});
```

This actually breaks the original request, because it treats `{ status: 'available' }` as a body param instead of a query param.

bennycode:
Thank you very much for your quick response. I actually discovered your "api" package by chance because I am the author of the [unofficial Coinbase Node.js SDK](https://github.com/bennycode/coinbase-pro-node) and I noticed that Coinbase moved their [API docs](https://docs.cloud.coinbase.com/sign-in-with-coinbase/docs) to your platform. That is how I discovered "api" because there was this example:

```ts
const sdk = require('api')('@coinbase-exchange/v1.0#qgumw1pl3iz4yut');

sdk['ExchangeRESTAPI_GetFees']()
```

I got curious how you resolve the API specs with a simple `require` statement and tried the example from your README which got me an error. Then I understood that you are fetching and parsing the OAS file. Very cool approach by the way! Given the fact that `fetch` is part of Node.js 18 your approach has a lot of potential for browsers & Node.js apps. I built a TypeScript API client generator (based on Swagger 2.0) myself in the past ([Swaxios](https://github.com/welovecoding/swaxios)) and I appreciate that you are putting a fresh spin on it. So let's get your "Hello, World" example fixed. 💪

By the way, can you explain to what URL the string "@coinbase-exchange/v1.0#qgumw1pl3iz4yut" resolves?

erunion (approved):

erunion:
> Another issue appears to be that as soon as you add the 2nd metadata argument like so:
> 
> ```js
> const sdk = require('api')('https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/json/petstore.json');
> 
> sdk.findPetsByStatus({ status: 'available' }, {}).then(async (res) => {
>   console.log(res);
> }).catch(err => {
>   console.error(err);
> });
> ```
> 
> This actually breaks the original request, because it treats `{ status: 'available' }` as a body param instead of a query param.

And this is happening because the method accessors accept either `metadata` or `body, metadata` so in this case because `metadata` is an empty object we think that `{status: available}` is the request body.

There isn't an easy way to prevent this without looking at each parameter in `body` and `metadata` to see where it is in the spec and how it should be delivered without heavily slowing down requests but we can easily filter out empty or undefined objects from here prior to making that "is this first parameter a body or is it actually metadata?" determination.

I'll ticket something

PR #460, opened 2022-06-29
Kanad testing [do not merge]

kanadgupta:
Attempt at loading the SDK into a little test (see `packages/api/kanad-test.js` on this branch).

To help reproduce the issue, here are the steps I took:
1. Ran the following command from the `packages/api` subdirectory to load some OAS from [this URL](https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/json/petstore.json) into the `petstore` alias:

```sh
node dist/bin.js install https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/json/petstore.json
```
2. Updated `packages/api/.api/apis/petstore/package.json` so its reference the `api` package points to its parent, rather than v4.5.1 and re-ran `npm i` in the `packages/api/.api/apis/petstore` directory[^1]

3. I've tried running `packages/api/kanad-test.js` with `node` and `ts-node` (and have tried a number of `require`/`import` patterns) and have run into some weird import error 😕 

[^1]: one weird thing I noticed with this: running `npm i` twice yields a weird `Cannot set property 'peer' of null` error.


packages/api/.api/api.json
```diff
@@ -0,0 +1,11 @@
+{
+  "version": "1.0",
+  "apis": [
+    {
+      "identifier": "petstore",
+      "source": "https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/json/petstore.json",
+      "integrity": "sha512-FSapRqUaxZYIPMPT+DN0E6xUfVdD/WadjTacNvwrk7scTzXhTZGNReEQ/lZfg9cz8yIvec2fqhTUAF2L4jVfRQ==",
+      "installerVersion": "4.2.0"
```
kanadgupta, line 8: I feel like this should be something else, but I don't know what to change it to 😬

packages/api/.api/apis/petstore/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "@api/petstore",
+  "main": "./index.ts",
+  "dependencies": {
+    "api": "../../../",
```
kanadgupta, line 5: This line is the only one I made changes to in this file.

packages/api/package-lock.json
```diff
@@ -61,6 +65,17 @@
...
       }
     },
+    ".api/apis/petstore": {
+      "name": "@api/petstore",
+      "dependencies": {
+        "api": "../../../",
+        "oas": "^18.3.3"
+      }
+    },
+    ".api/apis/petstore/node_modules/api": {
```
kanadgupta, line 75: I re-ran `npm i` in this directory after modifying the `package.json` in the `petstore` subdirectory and it looks like the link worked?

packages/api/kanad-test.js
```diff
@@ -0,0 +1,5 @@
+const SDK = require('@api/petstore/index');
```
kanadgupta, line 1: ESLint doesn't like this `require` pattern, but I've tried a bunch of different combinations that don't work 😕

packages/api/.api/api.json
erunion: Nah this is fine for now, it maps to the verison of `api` that you used when you did the install and because my branch is still at 4.2.0 that's what was used. `installerVersion` isn't actually used anywhere else yet, but I can see it used in the future where if we bump `api` and you have an old installer version we might be like "hey we made a bunch of performance improvements since <version>,  you should reinstall your SDK!"

packages/api/kanad-test.js
erunion: The overall ESLint config for our TS code is taking over here and wanting it to be an `import`. You can ignore ESLint here.

PR #482, opened 2022-07-12
feat: improved typescript method accessor `path` type generation

erunion:
## 🧰 Changes

While installing our OpenAPI definition to our homepage repository I noticed that all `get` accessors in the generated library had `path` typed to a simple `string` instead of the actual path in question. This fixes that!

With this work, autocompletion on HTTP method accessors should be a lot smarter and give you better inline docs:

screen_shot_2022-07-11_at_5 10 56_pm

## 🧬 QA & Testing

See tests.

### DIFF
--- a/packages/api/src/cli/codegen/languages/typescript.ts
+++ b/packages/api/src/cli/codegen/languages/typescript.ts
@@ -373,7 +373,7 @@ sdk.server('https://eu.api.example.com/v14');`)
     const parameters: OptionalKind<ParameterDeclarationStructure>[] = [{ name: 'path', type: 'string' }];
     const docblock: OptionalKind<JSDocStructure> = {
       description: writer => {
-        writer.writeLine(`Access any ${method} endpoint on your API.`);
+        writer.writeLine(`Access any ${method.toUpperCase()} endpoint on your API.`);
         return writer;
       },
       tags: [{ tagName: 'param', text: 'path API path to make a request against.' }],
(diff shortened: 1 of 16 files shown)


packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -589,7 +589,7 @@ sdk.server('https://eu.api.example.com/v14');`)
         this.methodGenerics.get(operation.method).addOverload({
           typeParameters,
           parameters: [
-            { name: 'path', type: 'string' },
+            { name: 'path', type: `'${operation.path}'` },
```
erunion, line 592: This is gross but if `type` is set to only `operation.path` then the generated TS output will be `get(path: /some/path)`, which obviously isn't valid. `type` here needs to literally be a string and yes `operation.path` is already a string and yes I also tried casting `operation.path` with `String(operation.path)`.

Just one of those weird `ts-morph` quirks.

packages/api/test/__fixtures__/definitions/operationid-quirks.json
```diff
@@ -13,7 +13,7 @@
     "/quirky-operationId": {
       "get": {
         "description": "This mess of a string is intentionally nasty so we can be sure that we're not including anything that wouldn't look right as an operationID for a potential method accessor in `api`.",
-        "operationId": "find/?*!@#$%^&*()-=_.,<>+[]{}\\|pets-by_status"
+        "operationId": "quirky/?*!@#$%^&*()-=_.,<>+[]{}\\|operation-id_string"
```
erunion, line 16: Changed this to be different because the `findPetsByStatus` operationID on a `/quirky-operationId` path was confusing the hell out of me.

packages/api/test/__fixtures__/sdk/simple-js-cjs/index.js
```diff
@@ -78,7 +78,7 @@ var SDK = /** @class */ (function () {
     this.core.setServer(url, variables);
   };
   /**
-   * Access any get endpoint on your API.
+   * Access any GET endpoint on your API.
```
domharrington, line 81: lol practicing what you preach here with the uppercase methods 😂

domharrington (approved):
👍

PR #469, opened 2022-07-08
docs: updating documentation for v5

erunion:
## 🧰 Changes

Updating all of our docs for the new v5 release.

All docs will be automatically deployed to https://api.readme.dev/ when changes are merged into `main` using the `rdme` workflow.

## 🧬 QA & Testing

It's probably easiest to QA and read these changes in the code viewer: https://github.com/readmeio/api/tree/docs/v5/packages/api

PR #485, opened 2022-07-13
chore: move ESLint/Prettier config to root, format docs

kanadgupta:
## 🧰 Changes

This sets up our ESLint and Prettier config at the root so we can properly format docs being created in #469.

- [x] Migrates all shared ESLint/Prettier packages and config to the root (the only thing I kept in place is the `.eslintrc` for the `api` subpackage since that's TS-specific, as well as the `.eslintrc` files for the test directories)
- [x] Made a slight enhancement to the `npm run prettier` command so it also formats Markdown files
- [x] Runs the `npm run prettier` command to format all of the Markdown files in this repo properly (minus `CHANGELOG.md`)
- [x] Set up an `alex` config
- [x] Added a check to the `linting` CI job to check our Markdown files for formatting and considerate language

## 🧬 QA & Testing

Does [the `linting` check in this PR](https://github.com/readmeio/api/runs/7328632915) yield the same 39 warnings as [the `linting` check](https://github.com/readmeio/api/runs/7306208227) in #469? And is [the new `prettier` check passing](https://github.com/readmeio/api/runs/7328632915?check_suite_focus=true#step:7:1)?


.github/CONTRIBUTING.md
```diff
@@ -37,7 +37,7 @@ Within `openapi.json` you'll want to spec out the OpenAPI definition that will b
 If you'd like to reference one of the definitions that we provide in the [`@readme/oas-examples`](https://npm.im/@readme/oas-examples) package then you can rename `openapi.json` to `openapi.js` and have a single line of the following and the test suite will pick it up!
 
 ```js
-module.exports = require("@readme/oas-examples/path/to/spec");
+module.exports = require('@readme/oas-examples/path/to/spec');
```
erunion, line 40: :nice-rgb:

.eslintignore
```diff
@@ -0,0 +1,9 @@
+# api subpackage
```
erunion, line 1: Wonder if these should be kept in their own package.

erunion (approved):

.eslintignore
kanadgupta: ESLint wasn't liking ignore files at the subpackage level 😕

PR #490, opened 2022-07-15
ci: getting started content reuse using hercule

### LINKED ISSUE #486: Content re-use for Getting Started page
Since [`getting-started.md`](https://github.com/readmeio/api/blob/42dbb7ad1ad79b89c5809423f653d96be776b9a7/docs/getting-started.md#L4) uses identical content to [README.md](https://github.com/readmeio/api/blob/42dbb7ad1ad79b89c5809423f653d96be776b9a7/packages/api/README.md), it'd be cool if we had a [little find-and-replace action](https://github.com/jacobtomlinson/gha-find-replace) during our syncing process, similar to [how we do in `rdme`](https://github.com/readmeio/rdme/blob/46cf29ed6da51652e079a28f1d0ca386b379da15/.github/workflows/docs.yml#L26-L41).

kanadgupta:
| 🚥 Fixes #486 |
| :-- |

## 🧰 Changes

This PR creates a lightweight docs build script using [`hercule`](https://github.com/jamesramsay/hercule) so we can reuse the content from `README.md` in `docs/getting-started.md`.

## 🧬 QA & Testing

I ran the action locally in a dry run state using `act` and confirmed it works. I also [did a dry run in an actual GitHub Actions runner](https://github.com/readmeio/api/runs/7364020684?check_suite_focus=true#step:5:1) (see https://github.com/readmeio/api/pull/490/commits/87155425ad0dde478ca4c2a7db67d92e1fdf09d9) and confirmed that that looked good as well:

image

As you can see above, the contents of `getting-started.md` will be updated since it will now have the TOC from `README.md` in it.

I also confirmed in the debug logs that the content of `getting-started` is being populated correctly:

image

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -3,44 +3,9 @@ title: Getting Started
 category: 5d4c940cae4e610060475769
 ---
 
-<p align="center">
-  <img width="400" src="https://raw.githubusercontent.com/readmeio/api/main/docs/images/logo.svg" />
-</p>
+<!--
+This doc is built using [hercule](https://github.com/jamesramsay/hercule)
+You can view the build script [here](../bin/build-markdown.js)
+-->
 
-<p align="center">
-  Magical SDK generation from an OpenAPI definition 🪄
-</p>
-
-<p align="center">
-  <a href="https://npm.im/api"><img src="https://img.shields.io/npm/v/api.svg?style=for-the-badge" alt="NPM Version"></a>
-  <a href="https://npm.im/api"><img src="https://img.shields.io/node/v/api.svg?style=for-the-badge" alt="Node Version"></a>
-  <a href="https://npm.im/api"><img src="https://img.shields.io/npm/l/api.svg?style=for-the-badge" alt="MIT License"></a>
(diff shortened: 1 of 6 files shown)


docs/getting-started.md
```diff
@@ -3,44 +3,9 @@ title: Getting Started
 category: 5d4c940cae4e610060475769
 ---
 
-<p align="center">
-  <img width="400" src="https://raw.githubusercontent.com/readmeio/api/main/docs/images/logo.svg" />
-</p>
+<!--
```
erunion, line 6: I wonder what rdmd does with comments in Markdown.

kanadgupta: It just doesn't render them out as HTML comments or anything at all, which is actually quite handy for internal dev notes like this. Take [this page](https://docs.readme.com/docs/rdme) for example. Here's some comments from its Markdown source: https://github.com/readmeio/rdme/blame/main/documentation/rdme.md#L7-L18

erunion: oh sweet

PR #494, opened 2022-07-21
fix: generated SDKs now ship an IFEE

erunion:
## 🧰 Changes

* [x] Updates our `type` and `interface` compilation to no longer export them due to a series of unreconcilable issues.
* [x] Updates our main default SDK export to now export an IFEE so people no longer need to do `new SDK()`, they just have a ready-to-use instance of the SDK.

### Before

```ts
import SDK from './packages/api/test__fixtures__/sdk/simple-ts';

const simpleSDK = new SDK();
simpleSDK.findPetsByStatus({ status: ['available'] }).then(res => {
  console.log(res)
});
```

### After
```ts
import simpleSDK from './packages/api/test__fixtures__/sdk/simple-ts';

simpleSDK.findPetsByStatus({ status: ['available'] }).then(res => {
  console.log(res)
});
```

## 🐳 Why were types and interfaces a problem?

1. [`ts-morph`](https://npm.im/ts-morph) doesn't have great, or any really, support for programmatically creating an IFEE. You can create an IFEE with `addVariableStatement` but when attempting to export it with `setIsDefaultExport` `ts-morph` will crash.
2. After managing to get an IFEE compiled into `export default createSDK` and `module.exports = createSDK`, the `.d.ts` TS declaration files would break with a TS2309 error because it would be exported as `export = _default`, which clashes with our JSON Schema types and interfaces that we're exporting as `export interface` and `export declare type`.
3. The [`json-schema-to-typescript`](https://npm.im/json-schema-to-typescript) library that we're using to convert JSON Schema schemas from an OpenAPI definition into TS code spits out a string dump containing multiple interfaces and types. We're able to handle this now by utilizing the `ts-morph` API to create a temporary source file, import our types + interfaces, and then export them as usable objects, but if we wanted to actually create them the proper way with `ts-morph` we can't because it doesn't have APIs for creating them dynamically. They expect you to know what you're setting into a type or interface, which we don't because it's coming from the API definition.

All of those reasons has led to me dropping support for exporting types and interfaces in our compiled TS declarations. I don't feel great about removing useful functionality here but believe me when I say that there's no other way that we can solve this right now.

We could maybe swap out `json-schema-to-typescript` with a custom JSON Schema compiler that sits directly on top of `ts-morph` but do I have interest in maintaining a JSON Schema compiler? lol. lmao. I'm also not even sure if properly creating, and exporting, these types and interfaces within `ts-morph` would even solve the problem we've got with this IFEE because `ts-morph` doesn't realy seem to like IFEE's very much.

That said, if someone really needs a type or interface exported they can export it themselves with the generated code that we create for them, they'll just need to re-add that if they generate it again. Not ideal, sure, but manageable.

## 🧬 QA & Testing

I've added new integration tests into the TS codegen suite for TS, CJS and ESM compiled SDKs. They all make a simple, mocked, API call. If they work, and TS doesn't complain about their types, then they will work everywhere.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -10,7 +10,7 @@ packages/api/.prettier*
 packages/api/CHANGELOG.md
 packages/api/dist/
 packages/api/example.js
-packages/api/test/
+packages/api/test/__fixtures/sdk/
 
 # httpsnippet-client-api subpackage
 packages/httpsnippet-client-api/.nyc_output/
--- a/.prettierignore
+++ b/.prettierignore
@@ -10,7 +10,7 @@ packages/api/.prettier*
 packages/api/CHANGELOG.md
 packages/api/dist/
 packages/api/example.js
-packages/api/test/__fixtures/sdk/
+packages/api/test/__fixtures__/sdk/
 
 # httpsnippet-client-api subpackage
 packages/httpsnippet-client-api/.nyc_output/
--- a/packages/api/src/cli/codegen/languages/typescript.ts
+++ b/packages/api/src/cli/codegen/languages/typescript.ts
@@ -3,6 +3,7 @@ import type { Operation } from 'oas';
 import type { HttpMethods, JSONSchema, SchemaObject } from 'oas/@types/rmoas.types';
 import type {
   ClassDeclaration,
+  FunctionDeclaration,
   JSDocStructure,
   MethodDeclaration,
   OptionalKind,
@@ -56,6 +57,8 @@ export default class TSGenerator extends CodeGeneratorLanguage {
 
   sdk: ClassDeclaration;
 
+  sdkExport: FunctionDeclaration;
+
   schemas: Map<
     string,
     {
@@ -97,9 +100,9 @@ export default class TSGenerator extends CodeGeneratorLanguage {
       },
       compilerOptions: {
         declaration: true,
+        outDir: 'dist',
         resolveJsonModule: true,
         target: options.compilerTarget === 'cjs' ? ScriptTarget.ES5 : ScriptTarget.ES2020,
-        outDir: 'dist',
 
         // If we're compiling to a CJS target then we need to include this compiler option
         // otherwise TS will attempt to load our `openapi.json` import with a `.default` property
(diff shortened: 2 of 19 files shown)


packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -186,23 +186,57 @@ export default class TSGenerator extends CodeGeneratorLanguage {
...
+       *
+       * Another quirk that this work avoids is there being an empty `export {};` at the very end
+       * of our compiled `d.ts` declaration file. I'm not sure why it was being added, and it
+       * didn't appear to be harming anything, but us manually creating this export statement
+       * causes it to go away.
+       *
+       * Thankfully, fortunately, and curiously, these are all only problems in non-CJS compiled
+       * targets. ¯\_(ツ)_/¯
+       */
+      sdkSource.addStatements('export default createSDK');
```
erunion, line 239: Absolutely seething at the amount of work I put in this work that boiled down to this one line.

.prettierignore
```diff
@@ -10,7 +10,7 @@ packages/api/.prettier*
 packages/api/CHANGELOG.md
 packages/api/dist/
 packages/api/example.js
-packages/api/test/
+packages/api/test/__fixtures__/sdk/
```
kanadgupta, line 13: my bad lol

packages/api/src/cli/codegen/languages/typescript.ts
kanadgupta, line 239: So wait, this still works fine in CJS right? This statement looks like ESM syntax 🤔

kanadgupta (approved):
not entirely following the code changes lol but the fixtures look good!

packages/api/src/cli/codegen/languages/typescript.ts
erunion: This is the TS/ESM side of things, the other bit on line 217 is what controls how its exported for CJS. TS takes the `createSDK` export assignment through `addExportAssignment` and creates `export = createSDK;`, which TS then compiles into `module.exports = createSDK`.

All this shit is so complicated.

PR #495, opened 2022-07-22
feat: adding support for case-insensitive header parameters

### LINKED ISSUE #178: Headers in snippets should be lowercased regardless of the source header
* [ ] Headers in snippets should be lowercased by default to make the snippet cleaner.
* [ ] Lowercased headers in metadata objects in `api` should be able to be properly matched to the API definition so you should be able to type in `accept`, `Accept`, `ACCEPT`, or `AcCePT` if you're wildnout.

erunion:
| 🚥 Fix #178 |
| :-- |

## 🧰 Changes

* [x] Updates snippet generation to present headers as lowercase.
* [x] Updates `api` parameter matching to be able to match case-insensitive parameters.

### DIFF
--- a/packages/httpsnippet-client-api/src/index.ts
+++ b/packages/httpsnippet-client-api/src/index.ts
@@ -1,6 +1,8 @@
 import type { Operation } from 'oas';
 import type { HttpMethods, OASDocument } from 'oas/@types/rmoas.types';
 import type { Client } from '@readme/httpsnippet/dist/targets/targets';
+import type { ReducedHelperObject } from '@readme/httpsnippet/dist/helpers/reducer';
+
 import stringifyObject from 'stringify-object';
 import { CodeBuilder } from '@readme/httpsnippet/dist/helpers/code-builder';
 import contentType from 'content-type';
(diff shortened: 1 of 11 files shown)


packages/httpsnippet-client-api/src/index.ts
```diff
@@ -88,7 +90,7 @@ const client: Client<APIOptions> = {
     link: 'https://npm.im/api',
     description: 'Automatic SDK generation from an OpenAPI definition.',
   },
-  convert: (source, options) => {
+  convert: ({ cookiesObj, headersObj, postData, queryObj, url, ...source }, options) => {
```
erunion, line 93: This minor refactor is to match how client targets are built now upstream in HTTPSnippet: https://github.com/readmeio/httpsnippet/blob/main/src/targets/node/fetch/client.ts#L24

domharrington (approved):
Idk if you've ever heard of this module: https://www.npmjs.com/package/caseless but it's what i've always used when i need to do case-insensitive object stuff. It was written specifically for Headers. It would also mean you wouldnt always have to lowercase, you could leave stuff as they're coming in to us as well?

erunion:
@domharrington Ah thanks, I forgot about `caseless`. I've swapped out some of the logic I did here with that.

PR #496, opened 2022-07-22
fix: removing undefined objects from incoming metadata parameters

### LINKED ISSUE #449: Filter out empty and undefined objects prior to making parameter placement determinations
https://github.com/readmeio/api/pull/447#issuecomment-1137676966

```js
const sdk = require('api')('https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/json/petstore.json');

sdk.findPetsByStatus({ status: 'available' }, {}).then(async (res) => {
  console.log(res);
}).catch(err => {
  console.error(err);
});
```

Making this request will fail because `status: available` is being treat as a request body because there's an (empty) `metadata` parameter being supplied.

* [ ] Filter out empty and undefined parameters from the method/operation accessors prior to running them through the `prepareParams` library.

erunion:
| 🚥 Fix #449 |
| :-- |

## 🧰 Changes

This updates our parameter determination work to filter out empty objects from the second `metadata` parameter that can be supplied to an SDK.

For example, given the following code:

```js
const sdk = require('api')('https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/json/petstore.json');

sdk.findPetsByStatus({ status: 'available' }, {}).then(async (res) => {
  console.log(res);
}).catch(err => {
  console.error(err);
});
```

When this is executed currently we treat `status: 'available'` as a body parameter because it's in the body parameter slot due to the empty object being in the metadata slot. Because `status: available` is actually metadata in this case, and we see that we have `metadata` present, when we do our "where should this data go" work happens we send `status` in the body payload. Problem is that this is a GET request so nothing actually gets sent.

The work here will in tune transform `findPetsByStatus({ status: 'available' }, {})` into `findPetsByStatus({ status: 'available' })` so that this metadata parameter will be sent as metadata (in this case a query parameter).

### DIFF
--- a/packages/api/test/core/prepareParams.test.ts
+++ b/packages/api/test/core/prepareParams.test.ts
@@ -358,6 +358,19 @@ describe('#prepareParams', function () {
       });
     });
 
+    it('should support matching on case-insensitive header parameters', async function () {
+      const operation = parameterStyle.operation('/anything/headers', 'get');
+      const metadata = {
+        PrimItive: 'buster',
+      };
+
+      expect(await prepareParams(operation, metadata)).to.deep.equal({
+        header: {
+          primitive: 'buster',
+        },
+      });
+    });
+
     it('should support path parameters', async function () {
       const operation = parameterStyle.operation('/anything/path/{primitive}/{array}/{object}', 'get');
       const metadata = {
--- a/packages/api/test/core/prepareParams.test.ts
+++ b/packages/api/test/core/prepareParams.test.ts
@@ -107,12 +107,11 @@ describe('#prepareParams', function () {
 
     const operation = spec.operation('/pet/findByStatus', 'get');
 
-    const body = {
+    const metadata = {
       status: ['available'],
     };
-    const metadata = {};
 
-    expect(await prepareParams(operation, body, metadata)).to.deep.equal({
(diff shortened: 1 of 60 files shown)


domharrington (approved):

packages/api/test/core/prepareParams.test.ts
```diff
@@ -101,6 +101,24 @@ describe('#prepareParams', function () {
...
+    await spec.dereference();
+
+    const operation = spec.operation('/pet/findByStatus', 'get');
+
+    const body = {
+      status: ['available'],
+    };
+    const metadata = {};
+
+    expect(await prepareParams(operation, body, metadata)).to.deep.equal({
```
Dashron, line 115: Which item here is supposed to be an empty object? from the name of the test I would assume `body` but it's not empty.

Dashron (changes requested):

packages/api/test/core/prepareParams.test.ts
erunion: @Dashron I've updated this to be `await prepareParams(operation, metadata, {})` to hopefully clear up any confusion here.

kanadgupta (approved):

PR #497, opened 2022-07-23
fix: bug where operationIDs with underscores were being modified

### LINKED ISSUE #488: XYZ does not appear to be a valid operation on this API
Hello, today I was trying to use your "api" package with this OAS file: https://dash.readme.com/api/v1/api-registry/qgumw1pl3iz4yut

When trying to use the SDK then I am always getting the following error message:

> Error: Sorry, `ExchangeRESTAPI_GetAccounts` does not appear to be a valid operation on this API.

According to the OAS file, there is a `operationId` with the value `ExchangeRESTAPI_GetAccounts`.

Am I holding it wrong?

```ts
import api from 'api';

// https://dash.readme.com/api/v1/api-registry/qgumw1pl3iz4yut
const sdk = api('@coinbase-exchange/v1.0#qgumw1pl3iz4yut');

// https://docs.cloud.coinbase.com/exchange/docs/authorization-and-authentication#creating-a-request
sdk.auth("cb-access-key", "API key as a string")
sdk.auth("cb-access-sign", "Passphrase for your the API key")
sdk.auth("cb-access-timestamp", "Timestamp for your request")
sdk.auth("cb-access-passphrase", "base64-encoded signature")

const accountId = 'some-id';

sdk['ExchangeRESTAPI_GetAccounts']({account_id: accountId})
  .then(console.log)
  .catch(console.error);
```

I am using **api v5.0.0-beta.1**.

It seems to work with **api v4.5.1** (although this gives me the error "Error: Sorry, this API currently requires multiple forms of authentication which we don't yet support.").

erunion:
| 🚥 Fix #488 |
| :-- |

## 🧰 Changes

This fixes a bug we had where if an operationID had underscores in it like `ExchangeRESTAPI_GetAccounts` our `oas` library would convert it into `ExchangeRESTAPIGetAccounts` because weren't recognizing underscores as being valid characters for a JS method accessor. This would result in calls to `sdk.ExchangeRESTAPI_GetAccounts` failing because it wasn't recognized as a valid operationID.

See https://github.com/readmeio/oas/pull/671 for the `oas` side of this fix.

## 🧬 QA & Testing

So we actually had some tests in place for this sort of thing but they were assuming a that underscores weren't valid. Egg on my face.

### DIFF
--- a/packages/api/src/cache.ts
+++ b/packages/api/src/cache.ts
@@ -1,4 +1,4 @@
-import type { OASDocument } from 'oas/@types/rmoas.types';
+import type { OASDocument } from 'oas/dist/rmoas.types';
 
 import 'isomorphic-fetch';
 import OpenAPIParser from '@readme/openapi-parser';
--- a/packages/api/test/index.test.ts
+++ b/packages/api/test/index.test.ts
@@ -109,7 +109,7 @@ describe('api', function () {
         // counterpart will.
         await petstoreSDK['find pet by id']()
           .then(() => {
-            throw new Error('This operationID should have thrown an exception');
+            throw new Error('This operationID accessor should have thrown an exception.');
           })
           .catch(err => {
             expect(err.message).to.equal(
--- a/packages/api/test/index.test.ts
+++ b/packages/api/test/index.test.ts
@@ -1,4 +1,4 @@
-import type { OASDocument } from 'oas/@types/rmoas.types';
+import type { OASDocument } from 'oas/dist/rmoas.types';
 
 import { assert, expect } from 'chai';
 import uniqueTempDir from 'unique-temp-dir';
@@ -10,8 +10,8 @@ import pkg from '../package.json';
 
 import { responses as mockResponses } from './helpers/fetch-mock';
 
-let petstoreSdk;
-let readmeSdk;
+let petstoreSDK;
+let readmeSDK;
 const petstoreServerUrl = 'http://petstore.swagger.io/api';
 
 describe('api', function () {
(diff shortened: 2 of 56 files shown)


packages/api/src/cache.ts
```diff
@@ -1,4 +1,4 @@
-import type { OASDocument } from 'oas/@types/rmoas.types';
```
erunion, line 1: All these `@types` changes are because I updated `oas` to store its declaration files in `dist/` instead of in a separate directory.

packages/api/test/index.test.ts
```diff
@@ -90,19 +90,34 @@ describe('api', function () {
...
     describe('#operationId()', function () {
       it('should work for operationId', async function () {
         fetchMock.get(`${petstoreServerUrl}/pets`, mockResponses.real('it worked!'));
 
-        expect(await petstoreSdk.findPets()).to.equal('it worked!');
+        expect(await petstoreSDK.findPets()).to.equal('it worked!');
       });
 
-      it('should work with operationIds that have contain spaces', function () {
-        expect(petstoreSdk['find pet by id']).to.be.a('function');
```
erunion, line 105: This test should have actually been failing but it wasn't because everything on `petstoreSDK` is a function... because that's how you call it. We only determine if `find by pet id` is actually an operationID until you call it, which we weren't doing here. Yikes!

packages/httpsnippet-client-api/test/__datasets__/operationid-with-underscores/output.js
```diff
@@ -1,5 +1,5 @@
 const sdk = require('api')('https://api.example.com/operationid-with-underscores.json');
 
-sdk.anythingOperation()
```
erunion, line 3: The `operationId` for this test is/was `anything_Operation`, which is totally fine as an SDK method.

domharrington (approved):

packages/api/src/cache.ts
```diff
@@ -1,4 +1,4 @@
...
+import type { OASDocument } from 'oas/dist/rmoas.types';
```
Dashron, line 1: Why are we importing from dist? Ideally we wouldn't have to.

Dashron (approved):

packages/api/src/cache.ts
erunion: This is the `main` and `types` entry for `oas`:

```
  "main": "dist/index.js",
  "types": "dist/index.d.ts",
```

I think we wouldn't need to load it from the dist if these types were exported from the main entrypoint.

Dashron: 👍

PR #508, opened 2022-08-02
fix: backport v5 operationID accessor changes

### LINKED ISSUE #507: operationId with dashes
We have an Open API Spec with a setting such as:

```json
"operationId":"get-api-token",
```

This generates a function such as:

```javascript
const sdk = require('api')('...');

sdk.auth('...', '...');
sdk.getApiToken({grant_type: 'client_credentials'}, {
  Accept: 'application/json',
  'Cache-Control': 'no-cache'
})
```

This works fine in the Readme.com reference documentation.  If you install the `api` library locally however it generates an error.

> Error: Sorry, `getApiToken` does not appear to be a valid operation on this API.

Editing the code to instead use `sdk['get-api-token']` works as expected:

```
< sdk.getApiToken({grant_type: 'client_credentials'}, {
---
> sdk['get-api-token']({grant_type: 'client_credentials'}, {
```

erunion:
| 🚥 Fix #507 |
| :-- |

## 🧰 Changes

When working on v5 I made a lot of changes to how code snippets are generated to cleanup operationID accessors that can't be cleanly used as a method accessor, like `get-a-pet` into `getAPet`. The problem is that I shipped these code snippet changes to production prematurely as our v5 work is still in an open beta.

This work here backports those changes to the v4.x release series so folks coming to `api` for the first time from a production code snippet.

## 🧬 QA & Testing

See the couple tests I added but to test this locally you can load into `packages/api` and run this code:

```js
const sdk = require('./src')({
  openapi: '3.0.0',
  info: {
    version: '1.0.0',
    title: 'Single Path',
  },
  servers: [
    {
      url: 'https://httpbin.org',
    },
  ],
  paths: {
    '/anything': {
      get: {
        operationId: 'get-pet',
        responses: {
          200: {
            description: 'OK',
          },
        },
      },
    },
  },
});

(async () => {
  await sdk.getPet().then(console.log);
  await sdk['get-pet']().then(console.log);
})();
```

You will see a screen like this:

Screen Shot 2022-08-02 at 2 36 26 PM

What you won't see are any "that operation ID could not be found" errors about the first `sdk.getPet()` call.

### DIFF
--- a/packages/api/src/index.js
+++ b/packages/api/src/index.js
@@ -20,16 +20,6 @@ class Sdk {
     this.cacheDir = opts.cacheDir ? opts.cacheDir : false;
   }
 
-  static getOperations(spec) {
-    return Object.keys(spec.api.paths)
-      .map(path => {
-        return Object.keys(spec.api.paths[path]).map(method => {
-          return spec.operation(path, method);
-        });
-      })
-      .reduce((prev, next) => prev.concat(next), []);
-  }
-
   load() {
     let authKeys = [];
     const cache = new Cache(this.uri, this.cacheDir);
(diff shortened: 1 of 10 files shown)


packages/api/src/index.js
```diff
@@ -70,30 +60,62 @@ class Sdk {
...
+          };
+
+          if (operationId !== originalOperationId) {
+            // If we cleaned up their operation ID into a friendly method accessor (`findPetById`
+            // versus `find pet by id`) we should still let them use the non-friendly version if
+            // they want.
+            op[originalOperationId] = ((operation, ...args) => {
+              return fetchOperation(spec, operation, ...args);
+            }).bind(null, next);
+          }
```
erunion, line 116: This work here is new but I'm going to add it to the v5 series as well. When I did the v5 work I completely dropped support for being able to do `sdk['get a pet']` if `get a pet` is the `operationID` in your spec, in favor of the always clean `sdk.getAPet()` accessor.

Problem is that anyone upgrading from v4 to v5 will now potentially see "get a pet" is not a valid operationID errors.

This double operationID work will only exist within the dynamic form of `api` as the SDK codegen work will always use the cleaner operationID (`getAPet`).

kanadgupta (approved):

erunion:
Published directly to v4.5.2 from here. https://github.com/readmeio/api/releases/tag/v4.5.2

PR #509, opened 2022-08-02
feat: allow original operationIDs to be used in dynamic SDKs

erunion:
## 🧰 Changes

This resolves a compatibility issue I introduced in v5 where because we now prefer cleaner, JS-method acessors to be used everywhere if you have an `operationId` that we deem to not be valid as a method accessor and clean it up, you can no longer use your original `operationId`.

For example if you have an operation that has `get-pet` as its `operationId` under v4 our code snippets would have you do this:

```js
await sdk['get-pet']().then(console.log);
```

In v5, and with our code snippets now in production we have you do this:

```js
await sdk.getPet().then(console.log);
```

If you upgrade from v4 to the v5 beta and have `sdk['get-pet']` we'll throw an error because though `get-pet` is in your spec we've rewritten it to be `getPet` and no longer recognize `get-pet`.

With this work here we'll now recognize both `sdk.getPet()` and `sdk['get-pet']()` as valid operationID accessors but will we'll continue to recommend you use `sdk.getPet()` as it's cleaner.

### DIFF
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -73,12 +73,28 @@ class Sdk {
           // cleans up any `operationId` that we might have into something that can be used as a
           // valid JS method.
           const operationId = next.getOperationId({ camelCase: true });
+          const originalOperationId = next.getOperationId();
 
-          return Object.assign(prev, {
+          const op = {
             [operationId]: ((operation: Operation, ...args: unknown[]) => {
               return core.fetchOperation(operation, ...args);
             }).bind(null, next),
-          });
+          };
+
+          if (operationId !== originalOperationId) {
+            // If we cleaned up their operation ID into a friendly method accessor (`findPetById`
+            // versus `find pet by id`) we should still let them use the non-friendly version if
+            // they want.
+            //
+            // This work is to maintain backwards compatibility with `api@4` and does not exist
+            // within our code generated SDKs -- those only allow the cleaner camelCase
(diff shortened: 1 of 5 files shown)


packages/api/src/index.ts
```diff
@@ -73,12 +73,28 @@ class Sdk {
...
+            // If we cleaned up their operation ID into a friendly method accessor (`findPetById`
+            // versus `find pet by id`) we should still let them use the non-friendly version if
+            // they want.
+            //
+            // This work is to maintain backwards compatibility with `api@4` and does not exist
+            // within our code generated SDKs -- those only allow the cleaner camelCase
+            // `operationId` to be used.
+            op[originalOperationId] = ((operation: Operation, ...args: unknown[]) => {
+              return core.fetchOperation(operation, ...args);
+            }).bind(null, next);
```
kanadgupta, line 94: wait so this code is only for the dynamic approach, correct?

kanadgupta (approved):

packages/api/src/index.ts
erunion: Correct `src/index` is the only part of this codebase that powers the dynamic approach.

erunion: Here's the work in the codegen code where it uses the camelCase version: https://github.com/readmeio/api/blob/main/packages/api/src/cli/codegen/languages/typescript.ts#L743-L750

PR #529, opened 2022-10-07
Fix example in README

maccman:
The original example was failing with a `could not determine executable to run` error.


erunion (approved):
ah good catch, thank you.

PR #530, opened 2022-10-10
fix: prefix schema titles that start with numbers to fix a type generation bug

### LINKED ISSUE #523: Issue with TS sdk generation: unclosed paren
Hello! I'm super excited about the ability to generate TS sdks from an OpenID definition.

I'm encountering an error when I run:

```sh
npx api install @journeyid/v1.0#2qy6513l3p6ue28 -l ts
```

That looks like this:

```sh
✖ Generating your SDK
Unexpected token, expected ")" (766:0)
   1 | import Oas from 'oas';
   2 | import APICore from 'api/dist/core';
   3 | import definition from './openapi.json';
   4 |
```

And indeed, something looks fishy around line 766:

```txt
   765 | type SystemExecutionsCreate_Response_2XX = (0
   766 | interface SystemPipelinesExecuteBodyParam {
   767 | pipelineKey: string
```

I have npm installed `"api": "^5.0.0-beta.3",`.

Let me know if you have any thoughts!

erunion:
| 🚥 Fix #523 |
| :-- |

## 🧰 Changes

A problem with `json-schema-to-typescript` was uncovered in https://github.com/readmeio/api/issues/523 where if a schema `title` prop starts with a number, that library will generate an invalid TypeScript interface causing codegen to fully crash out. To resolve this I've added a new `transformer` hook into [oas](https://npm.im/oas) in https://github.com/readmeio/oas/pull/695 to allow to adhoc transform these problematic to `json-schema-to-typescript` schemas into ones that can be used by prefixing them with an underscore. It's not a great solution but it's functional.

I've also upgraded all of our `oas`-related dependencies in the process as this change to `oas` to facilitate this new transformer incurred a breaking change there.

## 🧬 QA & Testing

I've added a test case for one of these schemas so all tests should be passing.

### DIFF
--- a/packages/api/src/cli/codegen/languages/typescript.ts
+++ b/packages/api/src/cli/codegen/languages/typescript.ts
@@ -44,6 +44,22 @@ function wordWrap(str: string, max = 88) {
   return str.replace(new RegExp(`(?![^\\n]{1,${max}}$)([^\\n]{1,${max}})\\s`, 'g'), '$1\n');
 }
 
+function schemaTransformer(schema: SchemaObject) {
+  if (schema.title) {
+    // If this schema starts with a number we should prefix it with something else as
+    // `json-schema-to-typescript` will generate an invalid TypeScript type or interface when
+    // this happens, resulting in a full codegen crash.
+    //
+    // https://github.com/readmeio/api/issues/499
+    if (/^\d/.test(schema.title)) {
+      // eslint-disable-next-line no-param-reassign
+      schema.title = `_${schema.title}`;
+    }
+  }
+
+  return schema;
+}
+
 export default class TSGenerator extends CodeGeneratorLanguage {
   project: Project;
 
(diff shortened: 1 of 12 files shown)


packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -508,7 +524,7 @@ sdk.server('https://eu.api.example.com/v14');`)
     paramTypes?: OperationTypeHousing['types']['params'],
     responseTypes?: OperationTypeHousing['types']['responses']
   ) {
-    const docblock: OptionalKind<JSDocStructure> = { tags: [] };
+    const docblock: OptionalKind<JSDocStructure> = {};
```
erunion, line 527: I made all these changes to docblock generation because I discovered while building this test spec for problematic schema titles that if an operation has no `summary` or `description` we would give it an empty `/** */` docblock.

kanadgupta (approved):

PR #533, opened 2022-10-13
refactor: json schema type generation and handling

erunion:
## 🧰 Changes

This completely refactors how we generate TypeScript types from JSON Schema by moving from [json-schema-to-typescript](https://npm.im/json-schema-to-typescript) to [json-schema-to-ts](https://npm.im/json-schema-to-ts). This allows us to do a number of cool new things:

* [x] Make the code that powers TS codegen a lot cleaner and easier to maintain. 
  * There's still a burden of entry into understanding everything that's going on, but there's less moving parts so it should be easier to pick up. 
* [x] Incredible speed improvements to SDK generation.[^1]
* [x] Removing a lot of nested Promise work that had the tendency to lock up the terminal when generating an SDK.
* [x] Move our generated types into new `schemas.ts` and `types.d.ts` files, cleaning up the main SDK file, and allowing these SDK types to be imported in folks' SDK implementations.
  * Without going into too much detail we couldn't do this before because `json-schema-to-typescript` generates a string of type exports and [ts-morph](https://npm.im/ts-morph) would generate bad code the way we had to insert them into the generated SDK. Now that we control how types are generated, we can safely use `ts-morph` to insert and export them.
* [x] General improvements to our codegen'd output to not generate unncessary `.d.ts` files for a TypeScript codegen'd SDK.
  * `.d.ts` files will only be generated if you're generating a JavaScript SDK.

## 🐳 Notes

I should note that by moving over to `json-schema-to-ts` we are losing the ability to have JSON Schema `description` content bubbled up into our types. This doesn't appear to be something that `json-schema-to-ts` supports, though I've filed an issue there asking on the best way to handle it (https://github.com/ThomasAribart/json-schema-to-ts/issues/92).

Though we are losing a vital piece of DX here I am comfortable with moving along without it as I think we can come up with a solution later to bring them back.

## 🌱 Demos

Here's what this new type generation looks like in action, and hey look they're all identical.

#### Typescript SDK
Screen Shot 2022-10-14 at 5 25 51 PM

#### JS SDK with a CJS export
Screen Shot 2022-10-14 at 5 25 56 PM

#### JS SDK with an ESM export
Screen Shot 2022-10-14 at 5 26 01 PM

[^1]: Petstore: 12,171ms → 2,816ms, ReadMe: 46,015ms → 4,523ms


packages/api/.eslintrc
```diff
@@ -11,12 +11,16 @@
...
+    // We use `@fixme` in some spots and it's fine.
+    "jsdoc/check-tag-names": "off",
+
     // Though we aren't requiring JSDoc blocks to be present, if they are they should be properly
     // formatted.
     "jsdoc/require-jsdoc": "off",
     "jsdoc/tag-lines": "off",
 
     // TypeScript types are better suited for these cases instead.
+    "jsdoc/require-param": "off",
```
erunion, line 23: Turning this off because we have some internal TS functions that have a docblock but because it's TS I don't really need to type out an unecessary `@param` JSDoc for them.

packages/api/package.json
```diff
@@ -51,13 +51,17 @@
...
+    "lodash.deburr": "^4.1.0",
     "lodash.merge": "^4.6.2",
+    "lodash.setwith": "^4.3.2",
+    "lodash.startcase": "^4.4.0",
     "make-dir": "^3.1.0",
-    "oas": "^19.0.0",
-    "object-hash": "^3.0.0",
+    "oas": "^19.0.1",
     "ora": "^5.4.1",
+    "prettier": "^2.7.1",
```
erunion, line 64: `json-schema-to-typescript` had Prettier as a dependency for formatting but since we aren't using that library anymore we need to now load in Prettier ourselves here so we can format the code we're generating. `ts-morph` by default doesn't generate code that looks great.

packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -137,17 +121,7 @@ export default class TSGenerator extends CodeGeneratorLanguage {
...
-
-  static formatter(content: string) {
-    return prettier(content, {
-      format: true,
-      style: {
-        printWidth: 120,
-        singleQuote: true,
-      },
-    } as JSONSchemaToTypescriptOptions);
+    this.schemas = {};
```
erunion, line 124: So I can do `JSON.stringify(schemas)` in `schemas.ts` and dump out all those schemas this is no longer a map because this is now being treated as a general object in order for me to set data into it with `lodash.setWIth()`.

packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -189,80 +163,114 @@ export default class TSGenerator extends CodeGeneratorLanguage {
...
+          isTypeOnly: true,
+          namedExports: types,
+          moduleSpecifier: './types',
+        },
+      ]);
+    } else {
+      // If we don't have any schemas then we shouldn't import a `types` file that doesn't exist.
+      sdkSource
+        .getImportDeclarations()
+        .find(id => id.getText() === "import type * as types from './types';")
```
erunion, line 192: Because we create the source file first, which does operation and response processing which in turn creates our schemas, we're inserting this import into the source file without having knowledge if we've actually got schemas for the spec we're using. This here is just cleaning that up so we aren't trying to import a file that we never actually made.

packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -189,80 +163,114 @@ export default class TSGenerator extends CodeGeneratorLanguage {
...
+             * There's an annoying quirk with `ts-morph` where if we're exporting a default export
+             * to a CJS environment, it'll export it as `exports.default`. Because we don't want
+             * folks in these environments to have to load their SDKs with
+             * `require('@api/sdk').default` we're overriding that here to change it to being the
+             * module exports.
+             *
+             * `ts-morph` unfortunately doesn't give us any options for programatically doing this
+             * so we need to resort to modifying the emitted JS code.
+             */
+            code = code.replace('exports.default = createSDK;', 'module.exports = createSDK;');
```
erunion, line 220: This is annoying but there's no way to do it otherwise with `ts-morph`.

packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -189,80 +163,114 @@ export default class TSGenerator extends CodeGeneratorLanguage {
...
-       * Thankfully, fortunately, and curiously, these are all only problems in non-CJS compiled
-       * targets. ¯\_(ツ)_/¯
-       */
-      sdkSource.addStatements('export default createSDK');
-    }
-
-    this.sdk.addProperties([
-      { name: 'spec', type: 'Oas' },
-      { name: 'core', type: 'APICore' },
-      { name: 'authKeys', type: '(number | string)[][]', initializer: '[]' },
```
erunion, line 263: Removed this because it wasn't being used anymore. I refactored all of the auth key stuff upstream into `api/core` but forgot to remove this then.

packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -272,21 +280,6 @@ export default class TSGenerator extends CodeGeneratorLanguage {
     });
 
     // Add our core API methods for controlling auth, servers, and various configurable abilities.
-    sdkSource.addInterface({
```
erunion, line 275: This interface exists in `api/core` so I can just load it for the SDK from that instead of having to duplicate it.

packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -383,85 +376,104 @@ sdk.server('https://eu.api.example.com/v14');`)
...
-       * However, that said, if somebody needs an interface or type exported they can export it
-       * themselves in the SDK code that we compile for them.
-       *
-       * @fixme
-       */
-      sdkSource.addStatements(
-        // All expressions coming out of `json-schema-to-typescript` are exported so by popping this
-        // off we'll just be inserting plain interfaces and types into the SDK source.
-        exp.substring('export '.length)
-      );
```
erunion, line 425: 195696780-22aa19d0-0d2a-40ef-b7e4-d6eab99e11f2

packages/api/.eslintrc
erunion, line 23: Turning this off because we have some internal TS functions that have a docblock but because it's TS I don't really need to type out an unecessary `@param` JSDoc for them.

packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -383,85 +376,103 @@ sdk.server('https://eu.api.example.com/v14');`)
...
+               * constant we're adding into the codegen'd schema file. As there's no way, not even
+               * using `eval()` in this case, to convert a string to a constant we're prefixing
+               * them with this so we can later remove it and rewrite the value to a literal.
+               * eg. `'Pet'` becomes `Pet`.
+               *
+               * And because our TypeScript type name generator properly ignores `:`, this is safe
+               * to prepend to all generated type names.
+               */
+              let str = JSON.stringify(schema);
+              str = str.replace(/"::convert::([a-zA-Z_$\\d]*)"/g, '$1');
```
erunion, line 430: I don't like this but it works. We do similar things in HTTPSnippet for turning some strings into literals for snippets. https://github.com/readmeio/httpsnippet/blob/main/src/targets/node/fetch/client.ts#L149-L151

packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -562,9 +569,7 @@ sdk.server('https://eu.api.example.com/v14');`)
 
         parameters.body = {
           name: 'body',
-          type: paramTypes.body
-            ? this.schemas.get(paramTypes.body).tsType
-            : this.schemas.get(paramTypes.formData).tsType,
+          type: paramTypes.body ? paramTypes.body : paramTypes.formData,
```
erunion, line 572: `paramTypes` contains an object containing the TypeScript types we've created for `body` or `formData` params. For example in our own codegen'd SDK:

```ts
/**
 * Returns the doc with this slug
 *
 * @summary Get doc
 */
getDoc(
  metadata: types.GetDocMetadataParam
): Promise<types.GetDocResponse401 | types.GetDocResponse403 | types.ErrorDocNotfound> {
  return this.core.fetch('/docs/{slug}', 'get', metadata);
}
```

` types.GetDocMetadataParam` would be what's stored in `paramTypes.metadata`.

packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -782,18 +745,6 @@ sdk.server('https://eu.api.example.com/v14');`)
       });
     });
 
-    // Run through and convert every schema we need to use into TS types.
-    await Promise.all(
```
erunion, line 786: This promise was the source of a lot of the operation blocking work that would often lock up your terminal when generating an SDK.

packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -804,14 +755,23 @@ sdk.server('https://eu.api.example.com/v14');`)
...
    *
-   * @param operation
-   * @param operationId
    */
   prepareParameterTypesForOperation(operation: Operation, operationId: string) {
     const schemas = operation.getParametersAsJSONSchema({
       mergeIntoBodyAndMetadata: true,
       retainDeprecatedProperties: true,
-      transformer: schemaTransformer,
+      transformer: (s: SchemaObject) => {
```
erunion, line 763: I don't like that I have this same function in `prepareResponseTypesForOperation` but because it's calling `this.addSchemToExport` I had trouble moving this function into a top-level `TSGenerator` function (i.e. changing this to `transformer: this.transformJSONSchema`) and not having it fail on calling `this.addSchemaToExport` with saying `this` was undefined.

I'm fine with the minor code duplication.

packages/api/src/cli/codegen/languages/typescript/util.ts
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Convert a string that might contain spaces or special characters to one that can safely be used
+ * as a TypeScript interface or enum name.
+ *
+ * This function has been adapted and slighty modified from `json-schema-to-typescript`.
+ *
+ * @license MIT
+ * @see {@link https://github.com/bcherny/json-schema-to-typescript}
+ */
+export function toSafeString(str: string) {
```
erunion, line 30: This is a safety net that I pulled over from `json-schema-to-typescript` to ensure that the types we're creating can be used in TS.

packages/api/src/cli/codegen/languages/typescript/util.ts
```diff
@@ -0,0 +1,74 @@
...
+  // identifiers in javaScript/ts:
+  // First character: a-zA-Z | _ | $
+  // Rest: a-zA-Z | _ | $ | 0-9
+
+  // remove accents, umlauts, ... by their basic latin letters
+  return (
+    deburr(str)
+      // if the string starts with a number, prefix it with character that typescript can accept
+      // https://github.com/bcherny/json-schema-to-typescript/issues/489
+      .replace(/^(\d){1}/, '$$1')
```
erunion, line 40: I added these lines to this function in order to fix I did in https://github.com/readmeio/api/pull/530 that resolved type names that started with a number from crashing codegen. If it starts with a number it'll get prefixed with a `$`, which is TS-safe.

packages/api/test/__fixtures__/sdk/operationid-quirks/index.d.ts
```diff
@@ -1,86 +0,0 @@
-import Oas from 'oas';
```
erunion, line 1: Because this `operationid-quirks` SDK fixture is TS-first, `index.d.ts` doesn't exist anymore because we don't generate `.d.ts` files for TS output (because TS can use the types in the `.ts` files instead).

packages/api/test/__fixtures__/sdk/petstore/index.ts
```diff
@@ -163,36 +180,48 @@ class SDK {
    *
    * @summary Finds Pets by status
    */
-  get(path: '/pet/findByStatus', metadata: FindPetsByStatusMetadataParam): Promise<FindPetsByStatus_Response_200>;
```
erunion, line 166: You'll notice that I've renamed these response types from `FinePetsByStatus_Response_200` to `FindPetsByStatusResponse200` as types should generally be pascal case.

packages/api/test/__fixtures__/sdk/readme/index.ts
```diff
@@ -78,113 +79,140 @@ class SDK {
    */
   get(
     path: '/api-registry/{uuid}',
-    metadata: GetAPIRegistryMetadataParam
-  ): Promise<GetAPIRegistry_Response_200 | Error_REGISTRY_NOTFOUND>;
```
erunion, line 82: `Error_REGISTRY_NOTFOUND` → `ErrorRegistryNotfound` is another example of us forcing type names to be pascal case.

packages/api/test/__fixtures__/sdk/readme/types.ts
```diff
@@ -0,0 +1,173 @@
+import type { FromSchema } from 'json-schema-to-ts';
```
erunion, line 1: You might (not) notice it but there are types being exported here that aren't actually used in the codegen'd SDK and that's because the way we generate JSON Schema in `oas` is to process all `components`. Because we're supplying the custom `transformer` to JSON Schema generation there in order to rewrite their names to `::convert::SCHEMA_NAME` so we can later turn it into a literal, this transformer work is running on all components regardless if they're being used or not.

I'm okay with this! Even if the type isn't used in the codegen'd SDK making it available to folks seems doesn't seem like a problem.

packages/api/test/__fixtures__/sdk/simple-js-esm/schemas.d.ts
```diff
@@ -0,0 +1,292 @@
+declare const ApiResponse: {
+  readonly type: 'object';
```
erunion, line 2: I hate this `readonly` stuff but this is what TS does when you generate a declaration file for an exported constant.

packages/api/test/cli/codegen/languages/typescript/utils.test.ts
```diff
@@ -0,0 +1,40 @@
...
+      expect(generateTypeName('a')).to.equal('A');
+      expect(generateTypeName('abc')).to.equal('Abc');
+      expect(generateTypeName('ABcd')).to.equal('ABcd');
+      expect(generateTypeName('$Abc_123')).to.equal('Abc123');
+      expect(generateTypeName('123_$Abc')).to.equal('$123Abc');
+      expect(generateTypeName('Abc-de-f')).to.equal('AbcDeF');
+    });
+
+    it('should generate a safe name from the string to constant work we have in codegen for $ref pointers', function () {
+      expect(generateTypeName('::convert::Pet')).to.equal('ConvertPet');
```
erunion, line 37: Just in case someone, somehow, has `::convert::Some String` as a component/schema name, we won't get tripped up on it when we later add `::convert::` to it for our type output.

packages/api/src/cli/codegen/languages/typescript.ts
erunion: > Fixing this is, to be frank, a fucking HARD problem for a couple reasons:

lol. bet.

packages/api/src/cli/codegen/languages/typescript/util.ts
kanadgupta: So wait this is an issue with `json-schema-to-ts` as well?

packages/api/src/cli/codegen/languages/typescript.ts
kanadgupta: this was bugging me a lot, so glad you were able to fix this 🥳

kanadgupta (approved):
I'm still not completely able to follow everything that's going on tbh, but it's far cleaner and easier to follow now. And looking through the test fixtures helped a ton. Very excited to get this out. LGTM 🚀

packages/api/src/cli/codegen/languages/typescript/util.ts
erunion: `json-schema-to-ts` doesn't create type names like `json-schema-to-typescript` did, it just gives you the ability to generate type shapes from JSON Schema. We still need to create the names ourselves.

packages/api/src/cli/codegen/languages/typescript/util.ts
```diff
@@ -0,0 +1,74 @@
...
+ * @license MIT
+ * @see {@link https://github.com/bcherny/json-schema-to-typescript}
+ */
+export function toSafeString(str: string) {
+  // identifiers in javaScript/ts:
+  // First character: a-zA-Z | _ | $
+  // Rest: a-zA-Z | _ | $ | 0-9
+
+  // remove accents, umlauts, ... by their basic latin letters
+  return (
```
gratcliff, line 36: 😓

gratcliff (approved):
Code looks good - I'm really not super familiar with this so there's not much I can touch on from a high level.

I appreciate the level of consolidation, pruning, and tests here.

PR #538, opened 2022-10-18
feat: allowing `accept` headers to always be sent through as metadata

### LINKED ISSUE #448: Allow Accept headers to be sent regardless if they're defined in the OAS
This should resolve the issue brought up in https://github.com/readmeio/api/pull/447. Basically if an `Accept` header isn't defined in a paths `parameters` block we currently don't allow you to supply it in the `metadata` argument. Because `Accept` headers defined in `parameters` should actually be ignored ([per the spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#fixed-fields-10)) we should instead do a couple things:

* [ ] If a path has `responses` (optional in OpenAPI 3.1) and `responses.*.content` present we should default to sending an `Accept` header for whichever media type is JSON-compatible.
* [ ] If no media type in the response content is JSON-compatible we should pick the first and send that.
  * [ ] We should also **not** do `res.json()` when returning the response as that'll throw JSON parsing errors.
* [ ] If an `Accept` header is present in `metadata`, allow it through regardless if its value matches a media type in `responses.*.content`.

erunion:
| 🚥 Fix #448 |
| :-- |

## 🧰 Changes

* [ ] Updates our parameter digestion and analysis to allow `Accept` headers to always be sent.
* [ ] Updates [@readme/oas-to-har](https://npm.im/@readme/oas-to-har) to fix a quirk where we might prefer non-JSON responses. 
  * https://github.com/readmeio/oas-to-har/pull/130
* [ ] Removing some dead code from our caching architecture.

## 🧬 QA & Testing

### DIFF
--- a/packages/api/src/cache.ts
+++ b/packages/api/src/cache.ts
@@ -5,7 +5,6 @@ import fs from 'fs';
 import os from 'os';
 import path from 'path';
 
-import OpenAPIParser from '@readme/openapi-parser';
 import findCacheDir from 'find-cache-dir';
 import 'isomorphic-fetch';
 import makeDir from 'make-dir';
@@ -103,28 +102,6 @@ export default class Cache {
     }
   }
 
-  static validate(json: any) {
-    if (json.swagger) {
-      throw new Error('Sorry, this module only supports OpenAPI definitions.');
-    }
-
-    // The `validate` method handles dereferencing for us.
-    return OpenAPIParser.validate(json, {
-      dereference: {
-        // If circular `$refs` are ignored they'll remain in the API definition as `$ref: String`.
-        // This allows us to not only do easy circular reference detection but also stringify and
-        // save dereferenced API definitions back into the cache directory.
-        circular: 'ignore',
-      },
-    }).catch(err => {
-      if (/is not a valid openapi definition/i.test(err.message)) {
-        throw new Error("Sorry, that doesn't look like a valid OpenAPI definition.");
-      }
-
-      throw err;
-    });
-  }
-
   isCached() {
     const cache = this.getCache();
     return cache && this.uriHash in cache;
@@ -171,10 +148,11 @@ export default class Cache {
   }
 
   async load() {
-    // If the class was supplied a raw object, just go ahead and bypass the caching system and
-    // return that.
+    // If the class was supplied a raw object we should still validate and make sure that it's
+    // dereferenced in  order for everything to function, but we shouldn't worry about saving it
+    // into the cache directory architecture.
     if (typeof this.uri === 'object') {
-      return this.uri;
+      return Fetcher.validate(this.uri);
     }
 
     return this.fetcher.load().then(async spec => this.save(spec));
--- a/packages/api/src/core/prepareParams.ts
+++ b/packages/api/src/core/prepareParams.ts
@@ -144,7 +144,6 @@ function processFile(
 export default async function prepareParams(operation: Operation, body?: unknown, metadata?: Record<string, unknown>) {
   let metadataIntersected = false;
   const digestedParameters = digestParameters(operation.getParameters());
-  const hasDigestedParams = !!Object.keys(digestedParameters).length;
   const jsonSchema = operation.getParametersAsJSONSchema();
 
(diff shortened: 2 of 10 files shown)


packages/api/src/core/prepareParams.ts
```diff
@@ -192,45 +185,42 @@ export default async function prepareParams(operation: Operation, body?: unknown
     } else if (typeof metadata === 'undefined') {
       // No metadata was explicitly provided so we need to analyze the body to determine if it's a
       // body or should be actually be treated as metadata.
-      if (!hasDigestedParams) {
```
erunion, line 195: Because `Accept` isn't defined as a traditional parameter in an OAS, it'll never show up in the `digestedParameters` object, and will never be here so if you sent an API request of **only** an `Accept` header, it would be treated as a body payload and sent that way.

packages/api/src/cache.ts
```diff
@@ -171,10 +148,11 @@ export default class Cache {
   }
 
   async load() {
-    // If the class was supplied a raw object, just go ahead and bypass the caching system and
-    // return that.
+    // If the class was supplied a raw object we should still validate and make sure that it's
+    // dereferenced in order for everything to function, but we shouldn't worry about saving it
+    // into the cache directory architecture.
     if (typeof this.uri === 'object') {
```
erunion, line 154: We don't, or document, that people use the dynamic `api` with a JSON object (we use this in some tests to avoid going through our caching architecture) but we weren't dereferencing them before. 😬

packages/api/src/core/prepareParams.ts
```diff
@@ -21,7 +21,6 @@ import getJSONSchemaDefaults from './getJSONSchemaDefaults';
  *
  * @see {@link https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#parameterObject}
  * @see {@link https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#parameterObject}
- * @param parameters
```
erunion, line 24: Removed these because they're useless with TS types.

packages/api/src/core/prepareParams.ts
```diff
@@ -336,7 +328,7 @@ export default async function prepareParams(operation: Operation, body?: unknown
           break;
         case 'header':
           params.header[paramName.toLowerCase()] = value;
-          delete metadata[paramName];
```
erunion, line 339: We weren't properly cleaning out case-insensitive headers from here before, potentially resulting in some edge cases where `multipart/form-data` requests might not be handled properly.

packages/api/src/cache.ts
```diff
@@ -103,28 +102,6 @@ export default class Cache {
     }
   }
 
-  static validate(json: any) {
```
erunion, line 106: Dead code.

packages/api/src/core/prepareParams.ts
```diff
@@ -192,45 +185,42 @@ export default async function prepareParams(operation: Operation, body?: unknown
...
+        } else if (headerParams.has(value)) {
+          return true;
         }
+
+        return false;
+      }).length;
+
+      if (intersection && intersection / Object.keys(body).length > 0.25) {
+        /* eslint-disable no-param-reassign */
+        // If more than 25% of the body intersects with the parameters that we've got on hand,
```
gratcliff, line 215: I have no idea what this means...

gratcliff (approved):
LGTM

packages/api/src/core/prepareParams.ts
erunion:
```js
const body = {
  // post payload
};

const metadata = {
 // query params
 // headers
};

sdk.getSomeOperation(body, metadata);
sdk.getSomeOperation(metadata);
sdk.getSomeOperation(body);
```

Because all three of these operation calls are valid we don't have any way to know if the first item in the call is a body payload or parameters so we do this analysis and intersection work of the first param against what parameters we know about in the spec. If more than 25% of the first parameter intersects with the parameters we know then we assume it's parameters.

Issue #432, opened 2022-04-17
Make time limit on api call


Oranjoose:
Occasionally an api call to a server will just hang indefinitely. I couldn't find in the documentation any way to supply a timeout in case the server hangs. Or am I wrong? Thank you

erunion:
We currently don't have a way to set a timeout on requests, but we'll add one in our upcoming v5 release. Figure we can piggyback off our existing `sdk.config()` architecture with a `timeout` property there.

```js
sdk.config({ timeout: 30 });
```

PR #539, opened 2022-10-19
feat: completely overhauling how we're handling response data

### LINKED ISSUE #536: Restructure the data we're returning so it's easier to access Response data
With the way `api` currently is if you want to access response status codes or headers you need to set `sdk.config({ parseResponse: false })` and then handle `res.json()` or `res.text()` yourself. Because we can do better than this we can return all this data back for the user instead:

```js
const petstore = require('@api/petstore');
const { data, status, headers, res } = await petstore.listPets();
console.log(`My pets name is ${data[0].name}!`);
```

* `data`: equates to the current default data we return in the SDK promise (the result of `res.json()` or `res.text()`
* `status`: Will be the content of `res.status`
* `headers`: Maps to `res.headers`
* `res`: This is the raw `Response` object that the `fetch` API returns.

---

* [ ] Update `api/core` to return data in this shape.
* [ ] Update codegen to return data in this shape, with correct typing.
* [ ] Update code snippet generation to return data in this shape.
* [ ] Write an upgrade doc as this will be a breaking change.

erunion:
| 🚥 Resolves #536 |
| :-- |

This **completely** refactors how we're handling returning data to always return an object containing response data, response status, response headers, and the raw [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object. This is obviously a breaking change but by doing this we can not only make it very easy to access data out of the response (something you couldn't do before), but also remove the overly complicated `parseResponse` config option.
Here's what a petstore request looked like originally where you needed to access the response `status` and data:

```js
const petstore = require('@api/petstore');
petstore.config({ parseResponse: false });

await petstore
  .listPets()
  .then(async (res) => {
    const data = await res.json();
    console.log(`[${res.status}] and my pets name is ${data[0].name}!`);
  });
```

And now:

```js
const petstore = require('@api/petstore');
const { data, status, headers, res } = await petstore.listPets();
console.log(`[${status}] and my pets name is ${data[0].name}!`);
```

Less mess to access data you might need? ✅

Cleaner code? ✅

✅? ✅

Additionally now with this data schema we're also now able to type the HTTP statuses that an operation might return, as seen in this ugly video I recorded:

https://user-images.githubusercontent.com/33762/196828483-3d561b9c-9dac-4725-ab7d-9bcd89123180.mov

And beyond that, we're now also able to do the same for responses that hit HTTP code that falls between 400 and 599 with the introduction of a new `FetchError` error handler! It contains all the same data as a regular request does: `data`, `status`, `headers`, and `res`.

```js
await petstore
  .deletePet({ id: petId })
  .catch(({ status, data }) => {
    // status = 404
    // data = 'Not Found'
  });
```

Or if you'd like to check for something other than `FetchError`:

```js
await petstore
  .deletePet({ id: petId })
  .catch(err => {
    if (!(err instanceof FetchError)) {
      throw err;
    }

    // err.status = 404
    // err.data = 'Not Found'
  });
```

## ℹ️ Removing HTTP method generics

Due to a lot of complexities and drawbacks with typing our typed HTTP method generic accessor overloads, all SDKs codegen'd with `api` or used dynamically will no longer support being able to make a request with a generic HTTP method (i.e. `sdk.delete('/pet/1234')`).

The reason behind this is that it it's incredibly difficult to coerce TypeScript into supporting a typed `sdk.get('/pet/1234', { ...metadata... })` method overload while at the same time also supporting a generic `sdk.get('/any/path')` type. Not only does TypeScript in this case think that you're trying to use the `/pet/1234` accessor, and asks you to supply metadata for a request that probably doesn't need it, but once you do add metadata for that request, TypeScript then fails and says that `/any/path` doesnt match `/pet/1234`.

Beyond this, it's also incredibly difficult to support these types of generic accessors with regard to how we handle path parameters. You would expect a `sdk.get('/pet/1234')` request for the `/pet/{id}` request to work just fine but because the `id` parameter is a required path parameter (all path parameters are), if you executed **only** `sdk.get('/pet/1234')`, TypeScript would complain about the required `metadata` type containing `id` not being present. At that point if you're doing `sdk.get('/pet/1234', { id: 1234 })` or `sdk.get('/pet/{petId}, { id: 1234 })` you're really better off using `sdk.getPet({ id: 1234 })` instead.

Beyond that, these generic HTTP method accessors aren't surfaced in the code snippets we generate and the only way to know about them is to read the docs as they aren't mentioned anywhere else.

So all this said, we're removing the generic `sdk[http method]()` accessors from all forms of `api`. All API integrations with this library will now be much more deliberate and codegen'd SDKs will have significantly less bloat.

We may add a generic `sdk.fetch()` method at some point down the line that wraps the `fetch` API, but honestly if you need to access an endpoint that doesn't exist in the OpenAPI definition you're using this library probably isn't for you (or at for that use-case).

## 🧰 Changes

* [x] Refactor how we're handling data that we return from operation calls into a new shape that contains the parsed data (from `res.json()` or `res.text()`), the HTTP status code, response headers, and the raw [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object that we got back from the `fetch` call.
* [x] Improve the TypeScript typing on operations in codegen to allow us to individually type each response shape with the documented HTTP code it's a part of.
* [x] Remove generic HTTP method accessors from all forms of the library: dynamic and codegen.
* [x] Running `description` and `summary` data that we add into docblocks during codegen into our word wrapper utility.
* [x] Updated snippet generation to surface this new request flow.
* [x] Updated snippet generation to no longer use generic HTTP method accessors.

## 🧬 QA & Testing

There's a lot going on here but it's all covered by the test suite.

I'm more than happy to walk through any parts if you've got questions.


packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -290,16 +295,8 @@ export default class TSGenerator extends CodeGeneratorLanguage {
...
-              ),
-            tags: [
-              { tagName: 'param', text: 'config Object of supported SDK options and toggles.' },
-              {
-                tagName: 'param',
-                text: 'config.parseResponse If responses are parsed according to its `Content-Type` header.',
-              },
-            ],
+              writer.writeLine(wordWrap('Optionally configure various options that the SDK allows.')),
+            tags: [{ tagName: 'param', text: 'config Object of supported SDK options and toggles.' }],
```
erunion, line 299: Though we aren't using `config.parseResponse` anymore I'm keeping the entire config system in place everywhere because I'll need it to support custom request timeouts in https://github.com/readmeio/api/issues/432.

packages/api/src/core/index.ts
```diff
@@ -7,19 +7,29 @@ import fetchHar from 'fetch-har';
...
 
-export interface ConfigOptions {
-  /**
-   * By default we parse the response based on the `Content-Type` header of the request. You can
-   * disable this functionality by negating this option.
-   */
-  parseResponse: boolean;
-}
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface ConfigOptions {}
```
erunion, line 18: Keeping this around because I'm going to need it when I start work on adding support for custom timeouts in https://github.com/readmeio/api/issues/432.

packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -587,20 +524,31 @@ sdk.server('https://eu.api.example.com/v14');`)
...
+          } else if (status.length === 3 && status.toUpperCase().endsWith('XX')) {
+            const statusPrefix = status.slice(0, 1);
+            if (!Number.isInteger(Number(statusPrefix))) {
+              // If this matches the `_XX` format, but it isn't `{number}XX` then we can't handle
+              // it and should instead fall back to treating it as an unknown number.
+              return `FetchResponse<number, ${responseType}>`;
+            }
+
+            this.usesHTTPMethodRangeInterface = true;
+            return `FetchResponse<HTTPMethodRange<${statusPrefix}00, ${statusPrefix}99>, ${responseType}>`;
```
kanadgupta, line 542: whoa TIL that we have this `HTTPMethodRange` type 🧭

packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -466,63 +460,6 @@ sdk.server('https://eu.api.example.com/v14');`)
     return sourceFile;
   }
 
-  /**
-   * Create a generic HTTP method accessor on the SDK.
-   *
-   */
-  createGenericMethodAccessor(method: string) {
```
kanadgupta, line 473: you really weren't kidding about how much code this cleans up lol

packages/api/test/__fixtures__/sdk/operationid-quirks/index.ts
```diff
@@ -72,38 +71,21 @@ class SDK {
...
   }
 
   /**
-   * This operation has an `operationId` with hypens yet it should still be accessible in the dynamic `api` flow.
+   * This operation has an `operationId` with hypens yet it should still be accessible in the
+   * dynamic `api` flow.
    *
    */
-  hyphenatedOperationId<T = unknown>(): Promise<T> {
+  hyphenatedOperationId(): Promise<FetchResponse<number, unknown>> {
```
kanadgupta, line 88: Why is the `data` type set to `unknown` in these fixtures but not in others?

packages/api/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "baseUrl": "./src",
     "declaration": true,
     "esModuleInterop": true,
-    "lib": ["dom", "es2020"],
+    "lib": ["dom", "dom.iterable", "es2020"],
```
kanadgupta, line 7: ?

packages/api/test/integration.test.ts
```diff
@@ -54,11 +54,11 @@ describe('integration tests', function () {
...
-      const res = await api(usptoSpec).post('/{dataset}/{version}/records', body, metadata);
-      expect(res.uri).to.equal('/anything/v1/oa_citations/records');
-      expect(res.requestBody).to.equal('criteria=propertyName%3Avalue');
-      expect(res.headers).to.have.deep.property('content-type', 'application/x-www-form-urlencoded');
-      expect(res.headers).to.have.a.customUserAgent;
+      const { data } = await api(usptoSpec).performSearch(body, metadata);
+      expect(data.uri).to.equal('/anything/v1/oa_citations/records');
+      expect(data.requestBody).to.equal('criteria=propertyName%3Avalue');
+      expect(data.headers).to.have.deep.property('content-type', 'application/x-www-form-urlencoded');
+      expect(data.headers).to.have.a.customUserAgent;
```
kanadgupta, line 61: I'm confused on how these tests pass 🧐 isn't `data` here the response body and shouldn't the headers be separate?
```suggestion
      const { data, headers } = await api(usptoSpec).performSearch(body, metadata);
      expect(data.uri).to.equal('/anything/v1/oa_citations/records'); // not sure what this should be
      expect(data).to.equal('criteria=propertyName%3Avalue');
      expect(headers).to.have.deep.property('content-type', 'application/x-www-form-urlencoded');
      expect(headers).to.have.a.customUserAgent;
```

packages/api/test/core/index.test.ts
```diff
@@ -48,26 +52,32 @@ describe('APICore', function () {
       const operation = readme.spec.operation('/changelogs/{slug}', 'put');
 
       readme.setServer('https://dash.readme.com/api/v1');
-      expect(await readme.fetchOperation(operation, body, { slug })).to.deep.equal({
-        uri: '/api/v1/changelogs/new-release',
-        requestBody: body,
+      await readme.fetchOperation(operation, body, { slug }).then(({ data }) => {
+        expect(data).to.deep.equal({
+          uri: '/api/v1/changelogs/new-release',
+          requestBody: body,
```
kanadgupta, line 58: wait i'm also confused here lol, this is the response body?

kanadgupta (commented):
Changes look great overall! Excited about removing the code cruft. Kinda confused on some of the response data in the tests though, see below

packages/api/src/core/errors/fetchError.ts
```diff
@@ -0,0 +1,31 @@
...
+  /** The content of the response. */
+  data: unknown;
+
+  /** The Headers of the response. */
+  headers: Headers;
+
+  /** The raw `Response` object. */
+  res: Response;
+
+  constructor(status: number, data: unknown, headers: Headers, res: Response) {
```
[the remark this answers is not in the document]
erunion: It can but because we're already parsing `res` to get the response data we don't want to parse it again.

packages/api/test/__fixtures__/sdk/operationid-quirks/index.ts
erunion: Because these operations don't have a `responses` property in their OAS -- which is no longer required with OpenAPI 3.1.

packages/api/tsconfig.json
erunion: I believe I added this because I'm doing some iteration on the `Headers` API in tests and TS doesn't think that `Headers.entries()` exists without this library being present.

packages/api/test/integration.test.ts
erunion: These tests are the way they are because I'm mocking the response to return data that was sent, this way I can assert that we're sending the right things.

kanadgupta: Ah I totally didn't think about the fact that mocking could be a thing here 🤦🏽

packages/api/test/core/index.test.ts
kanadgupta: [(ignore me, see above)](https://github.com/readmeio/api/pull/539#discussion_r1006931204)

kanadgupta (approved):

gratcliff (approved):

PR #542, opened 2022-10-24
feat: addition of a new `timeout` config option

### LINKED ISSUE #432: Make time limit on api call

erunion:
| 🚥 Resolves #432 |
| :-- |

## 🧰 Changes

This creates a new `timeout` config option for `sdk.config()` that allows you to supply a custom timeout number (in milliseconds), that when set will create a new [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) for the `fetch` request that happens.

Unfortunately because `AbortController` was introduced to Node in v15 we need to pull in the [node-abort-controller](https://www.npmjs.com/package/node-abort-controller) polyfill until Node 14 is EOL'd in April.

## 🧬 QA & Testing

I added some new tests for this to `config.test.ts` to check when a timeout is shorter than the amount of time it takes a request to respond, and the reverse.


packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -296,7 +296,13 @@ export default class TSGenerator extends CodeGeneratorLanguage {
           {
             description: writer =>
               writer.writeLine(wordWrap('Optionally configure various options that the SDK allows.')),
-            tags: [{ tagName: 'param', text: 'config Object of supported SDK options and toggles.' }],
+            tags: [
+              { tagName: 'param', text: 'config Object of supported SDK options and toggles.' },
+              {
+                tagName: 'param',
+                text: 'config.timeout Override the default `fetch` request timeout of 30 seconds (30000ms).',
```
kanadgupta, line 303: I like how the description below mentions that this value is in milliseconds. Wanna use that description here as well? 

https://github.com/readmeio/api/blob/8c7860b94292a046b883a464b2e0bc5be6582f74/packages/api/src/core/index.ts#L21-L22

packages/api/test/helpers/fetch-mock.ts
```diff
@@ -29,6 +29,12 @@ export const responses = {
     };
   },
 
+  delay: (res, delay) => {
+    return new Promise(resolve => {
+      setTimeout(() => resolve(res), delay);
+    });
```
kanadgupta, line 35: Can't wait until we actually can use natively promisified timeouts

kanadgupta (approved):
Small docs suggestion but otherwise LGTM! One overarching question — I recall that we had mentioned in the docs that the codegen worked in both browser and Node environments, is this `AbortController` polyfill going to create problems for our browser support?

erunion:
> I recall that we had mentioned in the docs that the codegen worked in both browser and Node environments, is this `AbortController` polyfill going to create problems for our browser support?

The `AbortController` is supported in all browsers now thankfully. https://developer.mozilla.org/en-US/docs/Web/API/AbortController

kellyjosephprice (approved):
Not familiar with the api code, but these changes look good.

packages/api/test/config.test.ts
```diff
@@ -0,0 +1,60 @@
...
+      sdk.config({ timeout: 100 });
+
+      await sdk
+        .deletePet({ petId })
+        .then(() => assert.fail())
+        .catch(err => {
+          expect(err.message).to.equal('The operation was aborted.');
+        });
+
+      fetchMock.restore();
```
dannobytes, line 45: nit: don't think you need these `fetchMock.restore()` in each individual test since you have it called in the `afterEach` function.

packages/api/test/config.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  // eslint-disable-next-line mocha/no-setup-in-describe
+  this.beforeAll(function () {
+    // Set a unique cache dir so these tests won't collide with other tests and we don't need to go
+    // through the trouble of mocking out the filesystem.
+    Cache.setCacheDir(uniqueTempDir());
+  });
+
+  describe('timeout', function () {
+    beforeEach(function () {
+      sdk = api(petstore as unknown as OASDocument);
```
dannobytes, line 26: Oh, these cumbersome TS conversions 😆

packages/api/test/config.test.ts
```diff
@@ -0,0 +1,60 @@
...
+      fetchMock.restore();
+    });
+
+    it('should override the default `fetch` timeout and return if request is quick', async function () {
+      fetchMock.delete(`http://petstore.swagger.io/v2/pet/${petId}`, mockResponses.delay(response, 100));
+
+      sdk.config({ timeout: 500 });
+
+      await sdk.deletePet({ petId }).then(({ data }) => {
+        expect(data).to.deep.equal(response);
```
dannobytes, line 54: curious what would happen if `sdk.deletePet` never resolved, e.g. rejected instead? I think this test would still pass and never hit this assertion.

maybe add a `.catch()` block that fails when rejected?

dannobytes (approved):
Left a few minor things, nothing blocking. LGTM!

packages/api/test/config.test.ts
erunion: There'd be an uncaught exception which would trigger a test failure in Mocha.

PR #543, opened 2022-10-25
fix: quirks with OR auth configurations not working right

### LINKED ISSUE #522: Multiple auth tokens were supplied
Hi there, when I pass in a username and password, I'm getting the following message: 

`Error: Multiple auth tokens were supplied for the auth on this endpoint, but only a single token is needed.`

Here is my code: ```sdk.auth(`${username}`, `${password}`)```

Any ideas?

erunion:
| 🚥 Fixes #522 |
| :-- |

## 🧰 Changes

This fixes a quirk in how we handle auth requirements on operations where if an operation allowed either OAuth2 (or a single token) auth, or Basic Auth, whether or not you supplied two forms of auth to `sdk.auth()` the OAuth 2 auth scheme would always get picked as the determined auth. How this would break down was that if you supplied `sdk.auth(username, password)`, signifying your intentions to use Basic Auth, we'd throw an error saying that the auth scheme (OAuth 2) that we found didn't need two auth tokens.

To resolve this I've mostly completely rewritten how we handle auth determination to better support cases where OR auth configurations may be present.

## 🧬 QA & Testing

I added a bunch of new tests and also some tests for the case brought up in #522.

The `prepareAuth` test also got a lot of changes so you might want to view the diff with whitespace changes turned off.


packages/api/test/dist.test.ts
```diff
@@ -34,8 +34,8 @@ describe('typescript dist verification', function () {
   });
 
   it('should be able to set an auth token', async function () {
-    const user = 'username';
-    const pass = 'changeme';
+    const user = 'buster';
```
erunion, line 37: Changed these from `username` everywhere so it's clearer that we have a real username in use (and not somehow turning a `username` variable into a string literal).

<sub>also :buster-chomp:</sub>

packages/api/src/core/prepareAuth.ts
```diff
@@ -45,40 +81,24 @@ export default function prepareAuth(authKey: (number | string)[], operation: Ope
...
-
       if (scheme.in === 'query' || scheme.in === 'header' || scheme.in === 'cookie') {
         preparedAuth[scheme._key] = authKey[0];
       }
       break;
 
     default:
-      throw new Error(`Sorry, this API currently supports a scheme, ${scheme.type}, that we don't yet support.`);
+      throw new Error(
+        `Sorry, this API currently supports a security scheme, ${scheme.type}, which this library doesn't yet support.`
```
kanadgupta, line 100: The double-usage of the word "supports" here felt a bit confusing:

```suggestion
        `Sorry, this API currently uses a security scheme, ${scheme.type}, which this library doesn't yet support.`
```

kanadgupta (approved):
Small error message copy suggestion but otherwise LGTM!

packages/api/src/core/prepareAuth.ts
```diff
@@ -18,25 +16,63 @@ export default function prepareAuth(authKey: (number | string)[], operation: Ope
...
+
+  // If we have two auth tokens present let's look for Basic Auth in their configuration.
+  if (authKey.length >= 2) {
+    // If this operation doesn't support HTTP Basic auth but we have two tokens, that's a paddlin.
+    if (!('Basic' in preparedSecurity)) {
+      throw new Error('Multiple auth tokens were supplied for this endpoint but only a single token is needed.');
+    }
+
+    const schemes = preparedSecurity.Basic.filter(s => usableSecuritySchemes.includes(s._key));
+    if (!schemes.length) {
```
RyanGWU82, line 53: I'm not entirely sure what you're trying to check for here but I don't think this matches with the error message on the next line.

packages/api/test/core/prepareAuth.test.ts
```diff
@@ -1,154 +1,226 @@
...
-  describe('HTTP', function () {
-    describe('scheme: basic', function () {
-      const user = 'username';
-      const pass = 'changeme';
+    describe('HTTP', function () {
+      describe('scheme: basic', function () {
+        const user = 'buster';
+        const pass = 'hunter1';
+
+        it('should supprot basic auth', function () {
```
RyanGWU82, line 92:
```suggestion
        it('should support basic auth', function () {
```

RyanGWU82 (approved):
Generally makes sense to me — the updates to `prepareAuth.ts` are pretty readable, although I'm not familiar with much about auth for this library. Test coverage helps a lot with trusting that it's working as intended.

packages/api/src/core/prepareAuth.ts
erunion: Yeah good catch, I've revised the error message to be clearer and added a comment here explaining what the hell I'm doing.

PR #546, opened 2022-10-25
feat: allowing users to manually specify an `authorization` header

erunion:
| 🚥 Fixes RM-4020 |
| :-- |

## 🧰 Changes

Like I did for `accept` headers in https://github.com/readmeio/api/pull/538 this allows users the ability to manually specify an `authorization` header in their API calls[^1], as such:

```js
const spec = require('./spec.json');
const sdk = require('api')(spec);

sdk.operation({limit: '10', page: '1', Authorization: '123'})
```

For the meat of the change that allowed this to happen check out https://github.com/readmeio/oas-to-har/pull/131.

## 🧠 Why would you want to do this?

I won't lie and say that the `sdk.auth()` handling is perfect because it definitely isn't. For example if you have an operation that accepts either OAuth2 or Basic Auth and you do `sdk.auth('buster')` for a Basic Auth, depending on how the security scheme is set up in the OAS and if OAuth2 is configured first we may attempt to make that API call as `Authorization: Bearer buster`. You can fix this by doing `sdk.auth('buster', '')` but allowing you to give the library an `Authorization` header is a good compromise in the "this should just goddamn work" category of things.

[^1]: If you specify an `authorization` header in an API call like above it will override whatever's set in `sdk.auth()`.


packages/api/src/core/prepareParams.ts
```diff
@@ -153,9 +157,30 @@ export default async function prepareParams(operation: Operation, body?: unknown
...
-      "You supplied metadata and/or body data for this operation but it doesn't have any documented parameters or request payloads. If you think this is an error please contact support for the API you're using."
-    );
+    let throwNoParamsError = true;
+
+    // If this operation doesn't have any parameters for us to transform to JSON Schema but they've
+    // sent us either an `Accept` or `Authorization` header (or both) we should let them do that.
+    // We should, however, only do this check for the `body` parameter as if they've sent this
+    // request both `body` and `metadata` we can reject it outright as the operation won't have any
+    // body data.
+    if (body !== undefined) {
```
erunion, line 167: These changes fixed an edgecase I overlooked in the `accept` work in https://github.com/readmeio/api/pull/538 where if you did `api.operation({ accept: 'something' })` on an operation that had no parameters we'd throw the "this operation doesn't have any params" error.

The same case applied to this new `authorization` handling.

packages/api/test/integration.test.ts
```diff
@@ -100,37 +101,93 @@ describe('integration tests', function () {
...
+        };
 
-      const { data } = await api(petstore as unknown as OASDocument).addPet(body, metadata);
-      expect(data.uri).to.equal('/v2/pet');
-      expect(data.requestBody).to.equal('{"id":1234,"name":"buster"}');
-      expect(data.headers).to.have.deep.property('accept', 'text/xml');
-      expect(data.headers).to.have.a.customUserAgent;
+        const sdk = api(petstore as unknown as OASDocument);
+
+        sdk.auth('buster');
```
kanadgupta, line 119: Should we throw a warning if the user passes both an `.auth` parameter as well as an `Authorization` header?

packages/api/src/core/prepareParams.ts
```diff
@@ -153,9 +157,30 @@ export default async function prepareParams(operation: Operation, body?: unknown
...
+      if (typeof body === 'object' && body !== null && !Array.isArray(body)) {
+        if (Object.keys(body).length <= 2) {
+          const bodyParams = caseless(body);
+
+          if (bodyParams.has('authorization') || bodyParams.has('accept')) {
```
kanadgupta, line 172: How about using your `specialHeaders` here? I just wrote this off the top of my head so it might require a bit of cleaning up

```suggestion
          if (specialHeaders.some(header => bodyParams.has(header))) {
```

kanadgupta (approved):
A few tiny questions but otherwise LGTM!

packages/api/test/integration.test.ts
erunion: We don't have a great way to throw any warnings outside of `console.*` and I wouldn't want to clutter up any integrations that come from this that aren't outright errors.

erunion: Especially because `sdk.auth()` can't be chained, maybe you have `sdk.auth()` somewhere up your stack but where you need to make a single call you need/want to specify an auth header.

packages/api/src/core/prepareParams.ts
erunion: oh brilliant, i was trying to think of a cleaner way to handle this last night.

PR #550, opened 2022-10-28
feat(testing): codegen smoketest suite + fixes

erunion:
| 🚥 Resolves RM-3726 |
| :-- |

## 🧰 Changes

Using the API specification dataset from [apis.guru](https://apis.guru/) this introduces a new smoketest suite that runs SDK codegen on 1.7k real-world APIs. It's a slow test but it works. And thanks it it uncovered a whole buffet of bugs, quirks, and annoyances:

### Quirks
* [x] If a spec has any `$ref` pointers after dereferencing, signaling to us that it contains circular references, we'll now throw an error saying we can't generate an SDK for that yet.
  - Supporting circular references isn't something I have the fortitude to support yet, nor is it something we adequately support with ReadMe, so I've ticketed it [here](https://github.com/readmeio/api/issues549) for now.
* [x] If a spec doesn't have any operations (an empty or missing `paths` object) then codegen will throw an error saying we can't support that spec because it doesn't have an SDK to codegen for.

### Fixes
* [x] If a `description` or `summary` within an operation contained `/*` or `*/` codegen would completely fail. We're now escaping these so that they won't break files.
* [x] If we were creating a `@summary` docblock tag (we do this if an operation has a `summary` and `description`) we weren't wrapping it at 88 characters like we do with all other docblock entries.
* [x] Fixed some quirks with the ways that `body` parameters on operation overloads were being set in cases that the schema applied to that body didn't allow `additionalProperties`. 
* [x] A bug that was introduced with the [removal of general HTTP method accessors](https://github.com/readmeio/api/pull/539) where if an operation didn't have an `operationId` set in the spec we weren't creating SDK accessors for it. Because we always generate TS-safe accessor names for operation IDs we're now always creating accessors.
* [x] If a spec had discriminator mappings we would think that it had circular refs and some weird shit would happen deep inside [oas](https://npm.im/oas) resulting in a crash during codegen.
* [x] If a JSON Schema object had the same name as a JS reserved keyword (like `delete` or `new`) we weren't sanitizing it, resulting in a crash. We now sanitize all variable names we're creating with a list of reserved JS/TS keywords I compiled.
  - This reserved keywords list also contains stuff like `global` or `window` as while it's not a reserved keyword it'd be bad if someone had that in a schema name and generated an SDK to use in a browser.

## 🧬 QA & Testing

🐑 Does the smoketest run in CI? And pass??


packages/api/.mocharc.json
```diff
@@ -8,6 +8,5 @@
   ],
   "watch-files": ["src/**/*.ts", "test/**/*.ts"],
   "recursive": true,
-  "reporter": "spec",
-  "timeout": 200000
```
erunion, line 12: Instead of setting a global timeout for tests I'm now only setting timeouts in the codegen and smoketest suites as they're the only things that can take a while.

packages/api/package.json
```diff
@@ -13,7 +13,8 @@
     "prebuild": "rm -rf dist/; npm run prebuild.packageConfig",
     "prebuild.packageConfig": "node -p \"'// This file is automatically updated by the build script.\\nexport const PACKAGE_NAME = \\'' + require('./package.json').name + '\\';\\nexport const PACKAGE_VERSION = \\'' + require('./package.json').version + '\\';'\" > src/packageInfo.ts; git add src/packageInfo.ts",
     "prepack": "npm run build",
-    "test": "nyc mocha \"test/**/*.test.ts\""
+    "test": "nyc mocha $(find test -name '*.test.ts' -not -path '*/smoketest.test.ts')",
```
erunion, line 16: Because the smoketest runs on 1k+ specs and takes a good amount of time I'm splitting it out because I only ever want to intentionally run the smoketest.

packages/api/package.json
```diff
@@ -90,6 +91,7 @@
     "mocha": "^10.0.0",
     "mock-require": "^3.0.3",
     "nyc": "^15.1.0",
+    "oas-normalize": "^7.1.0",
```
erunion, line 94: Loading this in to convert Swagger specs that we get back from https://apis.guru into OpenAPI specs that `api` can codegen from as `api` doesn't support Swagger right now (ever?).

packages/api/src/cli/codegen/language.ts
```diff
@@ -36,6 +36,22 @@ export default abstract class CodeGeneratorLanguage {
...
+     * This check is barbaric but there are a number of issues with how the `transformer` work we
+     * have in `oas` and in `.getParametersAsJSONSchema()` and `.getResponseAsJSONSchema()` that
+     * are fully crashing when attempting to codegen an SDK for an API definition that has a
+     * circular reference.
+     *
+     * In order to get v5 out the door we're not going to support this case initialy.
+     *
+     * @see {@link https://github.com/readmeio/api/issues/549}
+     */
+    if (JSON.stringify(spec.api).includes('"$ref":"#/')) {
```
erunion, line 50: The chances of this experiencing false positives where you have a `$ref` property somewhere that isn't a `$ref` is going to be nonexistent, but this check is barbaric nonetheless.

packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -672,14 +679,15 @@ sdk.server('https://eu.api.example.com/v14');`)
    */
   prepareParameterTypesForOperation(operation: Operation, operationId: string) {
     const schemas = operation.getParametersAsJSONSchema({
+      includeDiscriminatorMappingRefs: false,
```
erunion, line 682: I added this new option in https://github.com/readmeio/oas/pull/698. With it turned off any discriminator mapping that is a ref pointer won't automatically include the components for that ref pointer alongside `discriminator.mapping`. We normally have this happen because our form system in the API Explorer needs it to be colocated like that but I don't need that logic here, and having those components there was causing lots of problems with the `transformer` work I have a few lines down from here transforming dereferenced `$ref` schemas into primitive strings -- that we later use to correlate them to generated TS types.

Codegeneration is a goddamn hard problem and it's been a _week_.

packages/api/test/__fixtures__/sdk/alby/index.ts
```diff
@@ -0,0 +1,713 @@
+import type * as types from './types';
```
erunion, line 1: Added this new test fixture to test out the awful issues I uncovered with discriminators. I'd rather have that case explicitly tested here than hidden in the smoketest.

packages/api/test/__fixtures__/sdk/operationid-quirks/index.ts
```diff
@@ -80,6 +80,23 @@ class SDK {
     return this.core.fetch('/quirky-operationId', 'get');
   }
 
+  /**
+   * This operation doesn't have an `operationId` so we should create one.
+   *
+   */
+  getNoOperationId(): Promise<FetchResponse<number, unknown>> {
```
erunion, line 87: This wasn't being added due to the aforementioned bug with operations not having an `operationId` not being added into SDKs.

packages/api/test/__fixtures__/sdk/optional-payload/index.ts
```diff
@@ -91,7 +91,7 @@ class SDK {
    *
    */
   updatePetWithForm(
-    body?: types.UpdatePetWithFormFormDataParam,
+    body?: types.UpdatePetWithFormFormDataParam | types.UpdatePetWithFormMetadataParam,
```
erunion, line 94: s/o to @Dashron for the assist on this one

packages/api/test/__fixtures__/sdk/optional-payload/schemas.ts
```diff
@@ -1,4 +1,4 @@
-const updatePetWithForm = {
+const UpdatePetWithForm = {
```
erunion, line 1: With the work to ensure that variables and methods I'm creating aren't reserved I'm capitalizing these schema names here. They're only used within the `types.ts` for passing into [json-schema-to-ts](https://npm.im/json-schema-to-ts) so their name doesn't matter _that_ much, just as long as it doesn't break anything I'm happy.

packages/api/test/__fixtures__/sdk/petstore/schemas.ts
```diff
@@ -22,109 +22,13 @@ const Category = {
...
-      description: 'User Status',
-      minimum: -2147483648,
-      maximum: 2147483647,
-    },
-  },
-  title: 'User',
-  'x-readme-ref-name': 'User',
-} as const;
-const createUsersWithArrayInput = {
+const CreateUsersWithArrayInput = {
```
erunion, line 25: We alphabetize these schemas so with the name casing change the ordering changed.

packages/api/test/auth.test.ts
```diff
@@ -18,13 +17,14 @@ const user = 'username';
 const pass = 'changeme';
 
 describe('#auth()', function () {
-  this.beforeAll(function () {
```
erunion, line 21: I'm not really sure how this was working because Mocha doesn't document `beforeAll()` anywhere.

packages/api/test/auth.test.ts
```diff
@@ -18,13 +17,14 @@ const user = 'username';
...
+  before(function () {
     // Set a unique cache dir so these tests won't collide with other tests and we don't need to go
     // through the trouble of mocking out the filesystem.
     Cache.setCacheDir(uniqueTempDir());
   });
 
-  beforeEach(function () {
+  beforeEach(async function () {
+    const securityOas = await loadSpec('@readme/oas-examples/3.0/json/security.json');
```
erunion, line 27: I changed up all of these spec imports because `oas.dereference` has the tendency to sometimes update the reference of the spec that it's loading, causing these imports sometimes to import a the content of an already dereferenced spec from another test -- and then it itself dereferencing again causing all sorts of flakiness. `loadSpec` loads the spec and then clones the object with `JSON.stringify` and `JSON.parse`.

packages/api/test/cache-custom.test.ts
```diff
@@ -24,9 +25,13 @@ describe('cache (custom directory)', function () {
     await fs.rmdir(cacheDir, { recursive: true });
 
     Cache.setCacheDir(originalCacheDir);
+
+    fetchMock.restore();
```
erunion, line 29: Missed this somehow so if you can this test with `-w` and made changes it'd crash with an error saying that the mock it was trying to mock out in the test was already set up.

packages/api/test/datasets/refresh-dataset
```diff
@@ -0,0 +1,388 @@
+#! /usr/bin/env node
```
erunion, line 1: We use this same script and dataset in [@readme/openapi-parser](https://npm.im/@readme/openapi-parser) this one just has a lot more exclusions.

https://github.com/readmeio/openapi-parser/blob/main/bin/refresh-real-world-apis

.github/workflows/ci.yml
```diff
@@ -54,3 +54,15 @@ jobs:
 
       - name: Run tests
         run: npm test --ignore-scripts
+
+  smoketest:
```
erunion, line 58: I'm debating when this thing should run. Because it takes a while running it on every PR commit seems like overkill. Maybe it should only run on `main` and we'll (me) will just need to wait to tag releases until it passes there?

gratcliff (approved):
Do we need to worry about `constructor` namespaces, or are we not worried cause this is case-sensitive?

erunion:
> Do we need to worry about `constructor` namespaces, or are we not worried cause this is case-sensitive?

Yeah I'll add `constructor` to the list of banned names.

erunion:
Smoketest is currently failing on a couple specs that are circular so I'm excluding those from the dataset right now but also chunking up the smoketest into smaller runs so we only run a couple hundred specs in each job.

...cause right now the whole smoketest took over 2 hours. 😬

.github/workflows/ci.yml
```diff
@@ -54,3 +52,24 @@ jobs:
...
+    runs-on: ubuntu-latest
+    # if: ${{ github.actor != 'dependabot[bot]' }}
+    strategy:
+      matrix:
+        regex:
```
RyanGWU82: If I'm reading the docs correctly, this will run 4 different jobs in parallel, is that correct?

If that's right, then you actually should remove jobs 2 and 4, and just keep 1 and 3. Job 2 would run through _everything_ except `^az` and job 4 would run through _everything_ except `^go`. I think 1 and 3 alone would cover everything you want to cover.

(Also, if you want to speed things up, maybe spit the B-Z job into smaller chunks, so that more happens in parallel?)

.github/workflows/ci.yml
```diff
@@ -54,3 +52,24 @@ jobs:
 
       - name: Run tests
         run: npm test --ignore-scripts
+
+  smoketest:
+    runs-on: ubuntu-latest
+    # if: ${{ github.actor != 'dependabot[bot]' }}
+    strategy:
+      matrix:
+        regex: []
```
RyanGWU82: Why'd you comment this out entirely? I was going to say "I don't think this matrix is right" (I think you want to include only items 1 and 3, not 2 and 4) but curious why you're removing it entirely?

.github/workflows/ci.yml
```diff
@@ -54,3 +52,24 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+
+      - run: npm ci
+      - run: npm run build
+      - run: npm run test:smoke -- -- --regex="${{ matrix.regex }}"
```
RyanGWU82: Is this correct with two sets of `--`?

packages/api/test/cli/codegen/languages/typescript/utils.test.ts
```diff
@@ -1,25 +1,23 @@
...
 
-import { generateTypeName, wordWrap } from '../../../../../src/cli/codegen/languages/typescript/util';
+import { docblockEscape, generateTypeName, wordWrap } from '../../../../../src/cli/codegen/languages/typescript/util';
 
 describe('ts codegen utils', function () {
-  describe('#wordwrap', function () {
-    it('should not touch a short string', function () {
-      expect(wordWrap('This is a short sentence.')).to.equal('This is a short sentence.');
+  describe('#docblockEscape', function () {
+    it('should not a safe string', function () {
```
RyanGWU82, line 7:
```suggestion
    it('should not touch a safe string', function () {
```

RyanGWU82 (approved):
Generally LGTM, haven't reviewed any of this codegen code before so let me know if there's anything in particular you think we should look at in more depth.

.github/workflows/ci.yml
RyanGWU82: oops never mind, this is outdated :)

.github/workflows/ci.yml
erunion: I'm actually refactoring this to split it up into 15 different chunks based on the total number of specs in the dataset. Doing this regex thing is really messy.

.github/workflows/ci.yml
erunion: Unfortunately yeah. `npm run test:smoke` from the root runs `npm run test:smoke --workspace=packages/api` and I need to add the two `--` there to supply my data down into `smoketest.test.ts`. It's... not great.

packages/api/test/cli/codegen/languages/typescript/smoketest.test.ts
```diff
@@ -42,25 +63,25 @@ describe('typescript smoketest', function () {
...
-  }
+      const oas = Oas.init(spec);
+      await oas.dereference({ preserveRefAsJSONSchemaTitle: true });
 
-  // eslint-disable-next-line no-plusplus
-  for (let idx = START_AT_INDEX; idx < START_AT_INDEX + MAX_APIS_TO_TEST; idx++) {
-    if (!(idx in dataset)) {
-      return;
-    }
+      const ts = new TSGenerator(oas, './path/to/spec.json', name);
```
RyanGWU82, line 77: Is this path correct? If the path isn't needed in testing can you just use `undefined` or at least `./path/not/needed`?

packages/api/test/cli/codegen/languages/typescript/smoketest.test.ts
```diff
@@ -42,25 +63,25 @@ describe('typescript smoketest', function () {
...
-    // eslint-disable-next-line mocha/no-setup-in-describe
-    dataset.forEach(({ name, url }) => {
-      if (!new RegExp(datasetRegex).test(name)) {
-        return;
-      }
-
-      it(`should generate an SDK for \`${name}\``, smoketestAPI(url, name));
-    });
+  dataset.forEach(({ name, url }) => {
+    it.skip(`should generate an SDK for \`${name}\``, async function () {
```
RyanGWU82: Supposed to be skipped?

RyanGWU82 (approved):
Looks better than the regex approach, SGTM!

packages/api/test/cli/codegen/languages/typescript/smoketest.test.ts
erunion: Changed it to `./path/not/needed.json`. This path is what gets set to the `definition` import in the codegen'd SDKs, and since this smoektest doesn't actualy running an SDK, just checks if codegen worked or not, this isn't needed.

erunion:
THE SMOKETEST IS PASSING IN CI.

PR #553, opened 2022-10-29
docs: v5

erunion:
> **Note**
> 
> The smoketest is currently failing in this PR because some Google APIs are now vailing validation that weren't last week. I'm excluding them all from being run from the smoketest in https://github.com/readmeio/api/pull/554.

1515415098-afecdb0272d74c2336419e4753c2d444


bin/build-markdown.js
```diff
@@ -4,22 +4,18 @@ const path = require('path');
 // eslint-disable-next-line import/no-extraneous-dependencies
 const { transcludeFile } = require('hercule/promises');
 
-/**
- * all Markdown files located at within the `docs` directory
- * @type {Array.<String>}
- */
+// All Markdown files located at within the `docs` directory.
```
erunion, line 7:
```suggestion
// All Markdown files are located within the `docs` directory.
```

docs/making-requests.md
```diff
@@ -0,0 +1,76 @@
+---
+title: Making Requests
```
erunion, line 2: This page used to be `http-requests` but when this gets resynced when its merged into `main` I'll go into the Dash and delete the old `http-requests` page.

docs/upgrading-from-v4.md
```diff
@@ -0,0 +1,32 @@
...
+
+```js
+const petstore = require('api')('https://raw.githubusercontent.com/readmeio/oas-examples/main/3.0/json/petstore.json');
+
+petstore.listPets().then(({ data }) => {
+  console.log(`My pets name is ${data[0].name}!`);
+});
+```
+
+If you were using the `.config({ parseResponse: false })` option, that option has been removed in favor of this new resovled datashape where we return `data`, `status`, `headers`, and `res` to you. You can see documentation on those [here](https://api.readme.dev/docs/making-requests).
```
kanadgupta, line 30:
```suggestion
If you were using the `.config({ parseResponse: false })` option, that option has been removed in favor of this new resolved data shape where we return `data`, `status`, `headers`, and `res` to you. You can see documentation on those [here](https://api.readme.dev/docs/making-requests).
```

docs/making-requests.md
```diff
@@ -0,0 +1,76 @@
...
+category: 62cc6ce22b8b6601da6cb12d
+---
+
+If the API you're using doesn't have any documented [operation IDs](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#user-content-operationid), `api` will generate some for you to use.
+
+If you're using code generation these will be immediately available and to see in your generated library, however due to the nature of the [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) architecture in the dynamic version of the library this isn't the case. For these cases you can check out the documentation for the API you're using to see.
```
kanadgupta, line 8:
```suggestion
If you're using code generation, these will be immediately available to use in your generated library. However, due to the nature of the [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) architecture in the dynamic version of the library, this isn't the case for dynamically generated libraries. For these cases, you can check out the documentation for the API you're using.
```

docs/making-requests.md
```diff
@@ -0,0 +1,76 @@
...
+Here, the data returned in the promise from `listPets` is an object containing the following data:
+
+- `data`: The data that came back in the API request. Under the hood this is the result of `api` running `res.json()` or `res.text()` on the [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object from the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).
+- `status`: This is the HTTP status code of the response.
+- `headers`: The headers that came back in the response. This is an instance of the [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers) object.
+- `res`: This is the raw [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object we received from the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).
+
+## Error Handling
+
+For every response that is returned with an HTTP status code between 400 and 599 `api` will throw a custom error: `FetchError`. Identical to how responses are returned from successful requests, instances `FetchError` can be destructured to obtain the same data. For example:
```
kanadgupta, line 33:
```suggestion
For every response that is returned with an HTTP status code between 400 and 599, `api` will throw a custom error: `FetchError`. Identical to how responses are returned from successful requests, instances of `FetchError` can be destructured to obtain the same data. For example:
```

docs/making-requests.md
```diff
@@ -0,0 +1,76 @@
...
+
+```js
+await petstore.deletePet({ id: petId }).catch(({ status, data }) => {
+  // status = 404
+  // data = 'Not Found'
+});
+```
+
+Or, alternatively if you'd like to check for an error other than `FetchError` you can do that too:
```
kanadgupta, line 42: Would you want to give an example of what another kind of error might be? I don't know if this makes sense:
```suggestion
Alternatively if you'd like to check for an error other than `FetchError` (i.e., an error with the SDK itself), you can do that too:
```

docs/making-requests.md
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  // err.status = 404
+  // err.data = 'Not Found'
+});
+```
+
+## Request Timeouts
+
+By default the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) generally has a default timeout of 30 seconds. If you wish to configure this you can with the `.config()` method:
```
kanadgupta, line 57:
```suggestion
By default, the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) generally has a default timeout of 30 seconds. If you wish to configure this, you can with the `.config()` method:
```

docs/making-requests.md
```diff
@@ -0,0 +1,76 @@
...
+
+```js
+petstore.config({ timeout: 100 });
```
kanadgupta, line 60:
```suggestion
petstore.config({ timeout: 100 }); // 100 milliseconds
```

docs/making-requests.md
```diff
@@ -0,0 +1,76 @@
...
+  .deletePet({ id: petId })
+  .then(() => {
+    // goodbye, friend
+  })
+  .catch(err => {
+    // err.message = The operation was aborted.
+  });
+```
+
+By supplying a timeout to `.config({ timeout })`, in milliseconds, `api` will automatically initialize an [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) to handle timing out your request if it takes longer than you wish. Please note, however, that if you set a custom timeout the error that is thrown will be an instance of `AbortError`, **not** `FetchError`.
```
kanadgupta, line 72:
```suggestion
By supplying a timeout to `.config({ timeout })`, in milliseconds, `api` will automatically initialize an [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) to abort your request if it takes longer than your `timeout` value. Please note that if you set a custom timeout, the error that is thrown will be an instance of `AbortError`, **not** `FetchError`.
```

kanadgupta (approved):
Happy with the structure and content overall, most of my feedback is little touch-ups that you can pick and choose as needed.

docs/upgrading-from-v4.md
erunion: wow what happened here

PR #593, opened 2022-12-09
fix: resolving typing issues in JS codegen

### LINKED ISSUE #588: CJS/JS intellisense suggestions doesn't align with actual types
Node version: v16.14.0
npm version: 8.3.1

1. Create a new directory, run `npm init -y`
2. Install and set up using our spec, then select JS + CJS options in prompts:

```sh
npx api@latest install @developers/v2.0#452hs23laswqxqy
```

2. Create a new file called `index.js`, note that when you bring up intellisense[^1] when importing the package, it suggests using `.default`:

CleanShot 2022-12-06 at 09 59 34@2x

And if you keep using intellisense, you'll notice that `readme.default.` starts suggesting functions:

CleanShot 2022-12-06 at 10 01 06@2x

If you continue using intellisense to populate a function and then attempt to run it, you'll run into a `TypeError`:

CleanShot 2022-12-06 at 10 04 06@2x

The fix here is to remove the `.default` and it works as expected, but then Intellisense isn't nearly as helpful unless you know the exact method to use. Here's a full demo of the behavior:

https://user-images.githubusercontent.com/8854718/205963131-c6108585-37af-4670-b909-2eade8666abd.mp4

[^1]: <kbd>ctrl</kbd> + <kbd>space</kbd> for me, and also for [most keyboard shortcut configurations](https://code.visualstudio.com/docs/editor/intellisense#:~:text=You%20can%20trigger%20IntelliSense%20in,(.)%20in%20JavaScript).

erunion:
| 🚥 Fixes #588 |
| :---------------- |

## 🧰 Changes

* [x] Update codegen for CJS exports to have the main `createSDK` export set as `export =` in the `.d.ts` file. Having it as `export default createSDK` was causing issues where the typing was expecting you to to do `sdk.default.someOperation` even though the way you load the SDK doesn't require you to have to destructure it off `default`.
* [x] This also updates JS type generation to no longer export types in the main SDK entry point as it's causing issues with the affformentioned default export issue and also JS users can't import these types anyways. They'll still have access to them by way of TS intellisense but if you want to import these types you'll obviously need to use TS codegen or load them from `types.d.ts`. And yes this does work even though the file has a `.d.ts` extension:

Screen Shot 2022-12-09 at 1 01 46 PM

## 🧬 QA & Testing

This is what using a CJS SDK looks like now:

Screen Shot 2022-12-09 at 1 03 55 PM

What it looked like before (with `export default createSDK` in the `index.d.ts` file):

Screen Shot 2022-12-09 at 1 04 33 PM

### DIFF
--- a/packages/api/src/cli/codegen/languages/typescript.ts
+++ b/packages/api/src/cli/codegen/languages/typescript.ts
@@ -177,8 +177,8 @@ export default class TSGenerator extends CodeGeneratorLanguage {
        * exported individually because TS has no way right now of allowing us to do
        * `export type * from './types'` on a non-named entry.
        *
-       * Types in the main entry point are only being exported for TS ouputs as JS users won't be
-       * able to use these them and it clashes with the default SDK export present.
+       * Types in the main entry point are only being exported for TS outputs as JS users won't be
+       * able to use them and it clashes with the default SDK export present.
        *
        * @see {@link https://github.com/microsoft/TypeScript/issues/37238}
        * @see {@link https://github.com/readmeio/api/issues/588}
--- a/packages/api/src/cli/codegen/languages/typescript.ts
+++ b/packages/api/src/cli/codegen/languages/typescript.ts
@@ -172,22 +172,29 @@ export default class TSGenerator extends CodeGeneratorLanguage {
       this.createSchemasFile();
       this.createTypesFile();
 
-      // Export all of our available types so they can be used in SDK implementations.
-      //
-      // We're exporting all of the types individually because TS has no way right now of allowing
-      // us to do `export type * from './types'` on a non-named entry.
-      //
-      // https://github.com/microsoft/TypeScript/issues/37238
-      const types = Array.from(this.types.keys());
-      types.sort();
+      /**
+       * Export all of our available types so they can be used in SDK implementations. Types are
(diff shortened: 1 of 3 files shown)


packages/api/src/cli/codegen/languages/typescript.ts
```diff
@@ -172,22 +172,29 @@ export default class TSGenerator extends CodeGeneratorLanguage {
...
-          moduleSpecifier: './types',
-        },
-      ]);
+      /**
+       * Export all of our available types so they can be used in SDK implementations. Types are
+       * exported individually because TS has no way right now of allowing us to do
+       * `export type * from './types'` on a non-named entry.
+       *
+       * Types in the main entry point are only being exported for TS ouputs as JS users won't be
+       * able to use these them and it clashes with the default SDK export present.
```
kanadgupta, line 181:
```suggestion
       * Types in the main entry point are only being exported for TS outputs as JS users won't be
       * able to use them and it clashes with the default SDK export present.
```

erunion:
Ignore the smoketest failures, the dataset from https://api.apis.guru seems to delete specs when they get version bumps and we run with a local copy of https://api.apis.guru/v2/list.json

kanadgupta (approved):
other than the small comment suggestions, LGTM!

PR #605, opened 2023-01-14
fix: moving off `fs/promises` to `fs` in client-side code

### LINKED ISSUE #595: SDK not working in browsers
Hey, 
In the FAQ it says 
"If you generate an SDK with the CLI installation process then yes! If you're having trouble getting autogenerated SDKs working in a browser"

I use `npx api install ***` to install my sdk, but when I try to use it I get `Module not found: Can't resolve 'fs/promises'`
I see that the generated code imports `/api/dist/core` which import the `prepareParams.ts`, and that file imports nodejs modules. 

Thank you, for your help.

erunion:
| 🚥 Fixes #595 |
| :---------------- |

## 🧰 Changes

Some client frameworks seem to not polyfill `fs/promises` which we're using in our parameter preparation tooling for handling files being supplied to SDKs. As they **do** polyfill `fs`, I'm switching the bit where we use `fs/promises` to that and then writing an ESLint rule to prevent us from introducing a regression.


domharrington (approved):

kanadgupta (commented):
changes look good but I'm seeing a CI failure that might be related!

PR #606, opened 2023-01-17
Adds ESM usage instructions

ajmeese7:
## 🧰 Changes

Adds documentation for how to use the library with ESM syntax, which is something I struggled with for a bit.

## 🧬 QA & Testing

I've used this syntax on my end and it works identically to the `require` syntax.


erunion (changes requested):

erunion (approved):
thank you!

PR #660, opened 2023-06-27
fix(snippet): bug where path params may lose hyphens if they have one

erunion:
| 🚥 Fixes CX-380 |
| :---------------- |

## 🧰 Changes

This fixes a quirk in snippet generation where if a path param in an OpenAPI definition had a hyphen like `app-id` then the snippet that we'd generate for this would be `appid`. This would cause these requests in `api` code to fail because `appid` is not a parameter that exists.

### DIFF
--- a/packages/httpsnippet-client-api/src/index.ts
+++ b/packages/httpsnippet-client-api/src/index.ts
@@ -113,6 +113,7 @@ const client: Client<APIOptions> = {
 
     const operationSlugs = foundOperation.url.slugs;
     const operation = oas.operation(foundOperation.url.nonNormalizedPath, method);
+    const operationPathParameters = operation.getParameters().filter(param => param.in === 'path');
     const path = operation.path;
     const authData: string[] = [];
     const authSources = getAuthSources(operation);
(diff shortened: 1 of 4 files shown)


packages/httpsnippet-client-api/src/index.ts
```diff
@@ -169,7 +170,20 @@ const client: Client<APIOptions> = {
     Array.from(Object.entries(operationSlugs)).forEach(([param, value]) => {
       // The keys in `operationSlugs` will always be prefixed with a `:` in the `oas` library so
       // we can safely do this substring here without asserting this context.
-      metadata[param.substring(1)] = value;
+      const cleanedParam = param.substring(1);
+
+      // If our incoming path slug out of `oas.findOperation()` has been sanitized and is missing
+      // a hyphen, but there is a parameter in the OpenAPI definition that matches what our
+      // hyphen-less slug is then we should use that for the snippet.
+      const unsanitizedParam = operationPathParameters.find(p => {
```
erunion, line 178: This thing is/was only a problem on path parameters because we grab these params out of `oas.findOperation()` which uses the `path-to-regexp` library and munges our path params a bit.

PR #674, opened 2023-07-14
fix(codegen): don't surface error responses as method return types

### LINKED ISSUE #612: SDK generated by cli has incorrect return types
When generating an SDK via `npx api install <doc>`, the generated client will contain functions with return types composed of a union of all of the possible responses from a given API. For instance, take this one function generated by `npx api install @samsara-dev-rel/v2019.01.01#20op2bldf7yj03 -l ts`.

```ts
  createRoute(
    body: types.CreateRouteBodyParam
  ): Promise<
    | FetchResponse<200, types.CreateRouteResponse200>
    | FetchResponse<401, types.CreateRouteResponse401>
    | FetchResponse<404, types.CreateRouteResponse404>
    | FetchResponse<405, types.CreateRouteResponse405>
    | FetchResponse<429, types.CreateRouteResponse429>
    | FetchResponse<500, types.CreateRouteResponse500>
    | FetchResponse<501, types.CreateRouteResponse501>
    | FetchResponse<502, types.CreateRouteResponse502>
    | FetchResponse<503, types.CreateRouteResponse503>
    | FetchResponse<504, types.CreateRouteResponse504>
    | FetchResponse<number, types.CreateRouteResponseDefault>
  > {
    return this.core.fetch("/fleet/routes", "post", body)
  }
```

Of those, however, only `FetchResponse<200, types.CreateRouteResponse200>` is an actual possible return value. [As per the documentation, all of the response types from 400-599 will instead throw a FetchError](https://github.com/readmeio/api/blob/main/docs/making-requests.md#error-handling), and thus will never be returned by the function call.

The generated method should look more like this:

```ts
  createRoute(
    body: types.CreateRouteBodyParam
  ): Promise<
    | FetchResponse<200, types.CreateRouteResponse200>
    | FetchResponse<number, types.CreateRouteResponseDefault>
  > {
    return this.core.fetch("/fleet/routes", "post", body)
  }
```

erunion:
| 🚥 Fixes #612 |
| :---------------- |

## 🧰 Changes

This fixes a problem in codegen where we're documenting 400 and 500 status code families as being possible return types for SDK methods  which is not accurate as when we encounter these status codes we throw a `FetchError`. This is causing codegen's responses for these methods to effectively be typed as `unknown` as noted here[^1].

Unfortunately for us TypeScript doesn't offer any way to document that a method may throw an exception, and what that exception contains, so we need to fallback to a JSDoc `@throws` annotation. Not great but other than completely removing those error response types from the compiled SDK, which I don't want to do, we don't have any other option.

I have also cleaned up another very minor thing in this work with regard to typed method overloads that we're creating for methods that have an optional `body` or `metadata`. In these cases Because the first argument of the method can be either `body` or `metadata` we're creating a typed TS overload[^2] for each alternative. We were adding a docblock to each of these overloads but in actuality IDE Intellisense only ever picks up the docblock from the first method so adding the same docblock not only increases the size of the SDK we're generating but also fills the SDK with documentation that just isn't being used.

## 🧬 QA & Testing

The status code work is pretty well tested here with the SDK fixture snapshots we've got but for the docblock work you can see how the docblock from the first overload of this method is being picked up by Intellisense:

Screen Shot 2023-07-14 at 10 38 59 AM

For comparison, if we add that docblock to the last method it doesn't get picked up:

Screen Shot 2023-07-14 at 10 50 16 AM

[^1]: https://github.com/readmeio/api/issues/612#issuecomment-1633949991
[^2]: https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads

### DIFF
--- a/packages/api/src/cli/codegen/languages/typescript.ts
+++ b/packages/api/src/cli/codegen/languages/typescript.ts
@@ -3,7 +3,13 @@ import type { InstallerOptions } from '../language';
 import type Oas from 'oas';
 import type { Operation } from 'oas';
 import type { HttpMethods, SchemaObject } from 'oas/dist/rmoas.types';
-import type { ClassDeclaration, JSDocStructure, OptionalKind, ParameterDeclarationStructure } from 'ts-morph';
+import type {
+  ClassDeclaration,
+  JSDocStructure,
+  JSDocTagStructure,
+  OptionalKind,
+  ParameterDeclarationStructure,
+} from 'ts-morph';
 import type { PackageJson } from 'type-fest';
 
 import fs from 'fs';
@@ -28,7 +34,13 @@ interface OperationTypeHousing {
   operation: Operation;
   types: {
     params?: false | Record<'body' | 'formData' | 'metadata', string>;
-    responses?: Record<string, string>;
+    responses?: Record<
+      string | number,
+      {
+        description?: string;
+        type: string;
+      }
+    >;
   };
 }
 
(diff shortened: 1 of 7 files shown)


domharrington (approved):
LGTM!

Dashron (approved):

PR #675, opened 2023-07-15
feat(cli): adding a new `--identifier` option for supplying an SDK identifier

### LINKED ISSUE #608: Add identifier as a cli parameter
[extendable code in the codebase](https://github.com/readmeio/api/blob/main/packages/api/src/cli/commands/install.ts#L71)

## my use-case
Some of the openapi docs uploaded to readme.io are not consumable by default. So it would be nice to download a json definition as is, and manually fix the problems, and regenerate them time-to-time supporting the dev process more.

## what I tried
I wrote a command `rimraf ./.api && npx api install -l ts ./files/fixed-opensea-api.json` which is almost good. The problem with it that I need to type "opensea" every time I run this command.

## expected behaviour
`rimraf ./.api && npx api install -l ts -i opensea --install-missing ./files/fixed-opensea-api.json` would generate the types without any user interactions.

Probably [this part](https://github.com/readmeio/api/blob/main/packages/api/src/cli/commands/install.ts#L158) should be skipped too with a flag.

erunion:
| 🚥 Fixes #608 |
| :---------------- |

## 🧰 Changes

This adds a new optional `--identifier` option to the CLI interface for presetting the identifier you want to use with your generated SDK.

## 🧬 QA & Testing

This CLI has no unit tests at the moment so I just tested this out manually; notice I was not prompted to enter in an identifier.

Screen Shot 2023-07-14 at 4 57 28 PM

And my `debug` SDK was installed into `.api/apis` and will be able to be used as `@api/debug`:

Screen Shot 2023-07-14 at 4 57 47 PM

### DIFF
--- a/packages/api/src/cli/commands/install.ts
+++ b/packages/api/src/cli/commands/install.ts
@@ -4,7 +4,6 @@ import { Command, Option } from 'commander';
 import figures from 'figures';
 import Oas from 'oas';
 import ora from 'ora';
-import validateNPMPackageName from 'validate-npm-package-name';
 
 import Fetcher from '../../fetcher';
 import codegen from '../codegen';
@@ -18,6 +17,7 @@ cmd
   .name('install')
   .description('install an API SDK into your codebase')
   .argument('<uri>', 'an API to install')
+  .option('-i, --identifier <identifier>', 'API identifier (eg. `@api/petstore`)')
   .addOption(
     new Option('-l, --lang <language>', 'SDK language').choices([
       'js', // User generally wants JS, we'll prompt if they want CJS or ESM files.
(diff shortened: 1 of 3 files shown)


domharrington (approved):
LGTM! Could we prefill the identifier from `slug(info.api.title)` (or something else?) if they dont provide one? Could we also not prompt them if we detect they're in an environment where we can't accept user input (like CI, or automated installer)?

erunion:
> Could we prefill the identifier from `slug(info.api.title)` (or something else?) if they dont provide one?

Good idea. https://github.com/readmeio/api/issues/676

> Could we also not prompt them if we detect they're in an environment where we can't accept user input (like CI, or automated installer)?

Also a good idea. https://github.com/readmeio/api/issues/677

PR #688, opened 2023-08-04
refactor: removing code formatting + upgrading to prettier 3

erunion:
## 🧰 Changes

This upgrades our codebase to utilize Prettier 3. Unfortunately because Prettier 3 is now an ESM project, and though they still export CJS, I am having extreme troubles getting it working in our stack so I'm opting to remove it from our codegen SDKs where it was auto-formatting all code we generate.[^1]

I would love to bring this back at some point because I think there's value in having a generated SDK that *looks* nice, but because these SDKs are still functional it's fine to remove it for now.

[^1]: Having `prettier@3` as a dependency in `api` was also causing problems with projects that are still using `prettier@2` as NPM didn't like to have them installed alongside each other.


packages/api/test/__fixtures__/sdk/petstore/index.ts
```diff
@@ -1,5 +1,5 @@
 import type * as types from './types';
-import type { ConfigOptions, FetchResponse } from 'api/dist/core';
+import type { ConfigOptions, FetchResponse } from 'api/dist/core'
```
erunion, line 2: It's kinda funky that these type imports are no longer getting codegenerated with a `;` but it seems like they do if there's three types being imported here. This might be resolved in the latest `ts-morph` releases, which we're a couple majors behind.

But since this code still works I'm not going to fret over it now.

package.json
```diff
@@ -6,7 +6,8 @@
     "build": "npx lerna run build --stream",
     "build-docs": "node bin/build-markdown.js",
     "clean": "npx lerna clean",
-    "lint": "eslint . --ext .js,.ts",
+    "lint": "eslint . --ext .js,.ts; npm run lint:types",
+    "lint:types": "npm run lint:types --if-present --workspaces",
     "prepare": "husky install",
     "prettier": "prettier --list-different \"./**/**.{js,ts,md}\"",
     "prettier:write": "prettier --list-different --write \"./**/**.{js,ts,md}\"",
```
kanadgupta, line 13:
```suggestion
    "prettier": "prettier --check .",
    "prettier:write": "prettier --check --write .",
```

package.json
```diff
@@ -6,7 +6,8 @@
     "build": "npx lerna run build --stream",
     "build-docs": "node bin/build-markdown.js",
     "clean": "npx lerna clean",
-    "lint": "eslint . --ext .js,.ts",
+    "lint": "eslint . --ext .js,.ts; npm run lint:types",
```
kanadgupta, line 9:
```suggestion
    "lint": "eslint . --ext .js,.ts; npm run lint:types; npm run prettier",
```

kanadgupta (approved):
tiny prettier CI suggestion below, otherwise LGTM!

PR #699, opened 2023-08-18
fix: various quirks with how we handle `accept` headers

### LINKED ISSUE #697: Unable to Generate Token using api Package due to Schema Mismatch
I have been attempting to integrate with the Basiq API (v3) by following the documentation provided at https://api.basiq.io/reference/posttoken. My goal is to generate a new token. While I was successful in achieving this using tools like Postman and curl, I encountered consistent errors when trying to implement the same functionality using the provided Node.js example with the api package.

The primary issue is that regardless of the data I provided in the request body, I received errors indicating that the body was missing. This is perplexing since I ensured that the request body was formatted according to the documentation.

Upon closer investigation, I discovered that the issue lies within the api package's implementation, specifically in the prepareParams function located at [https://github.com/readmeio/api/blob/main/packages/api/src/core/prepareParams.ts#L269](https://github.com/readmeio/api/blob/main/packages/api/src/core/prepareParams.ts#L269). This function expects a schema entry with the type `body`. However, for the `/token` endpoint, the content type is set to `application/x-www-form-urlencoded`, resulting in the schema being generated with `formData` instead of body. Consequently, the body information is being removed from the request.

Interestingly, this problem seems to be isolated to the `/token` endpoint. Other endpoints function as expected, using the `application/json` content type and interacting well with the api package.

# Steps to Reproduce:

Attempt to generate a new token using the provided Node.js example and the api package for the `/token` endpoint.
Observe that the request fails, with errors indicating that the request body is missing.

# Expected Behavior:

The api package should handle requests to the `/token` endpoint seamlessly, despite the differing content type of `application/x-www-form-urlencoded`.

# Possible Workarounds:

Since the api package seems to struggle with the formData schema of the /token endpoint, you might consider the following workarounds:

**Directly using axios or node-fetch**: Bypass the api package and make requests to the /token endpoint using a library like axios or node-fetch.

**Adapting the api Package**: Modify the api package's prepareParams function to accommodate the formData schema in the /token endpoint. This might involve altering how the function interprets the schema type or enhancing its flexibility.

# Impact:

This issue prevents developers from effectively using the provided api package for token generation using the /token endpoint. It might lead to frustration and confusion when attempting to integrate with the Basiq API.

# Additional Information:

**Basiq API Documentation**: https://api.basiq.io/reference/posttoken
**Affected Endpoint**: `/token`
**Affected Package**: api (https://github.com/readmeio/api)

Please let me know if you need any further information or clarification on this issue.

OPENAPI Spec: [https://raw.githubusercontent.com/basiqio/api-ref/main/core.yml](https://raw.githubusercontent.com/basiqio/api-ref/main/core.yml
)

erunion:
| 🚥 Resolves #697 |
| :------------------- |

## 🧰 Changes

* [x] If we have an `accept` header in a HAR, and that header is JSON-like, we'll now stop adding it to the code snippet. This'll help us avoid unnecessary bloat, like in the following snippet, because `api` understands and prioritizes JSON over anything else.

```js
const sdk = require('api')('@basiq/v3.0#edsvz2rlk7p46u2');

sdk.auth('Basic someAuthToken');
sdk.postToken({scope: 'CLIENT_ACCESS'}, {accept: 'application/json', 'basiq-version': '3.0'})
```

* [x] If someone *does* supply an `accept` header into a code snippet, like the above one, we'll now always send that as a proper header instead of it *sometimes* getting lumped in with the body payload.
* [x] Fixed a quirk where a header could also be duped and send as a header **and** in a body payload.

### DIFF
--- a/packages/api/test/__fixtures__/definitions/basiq.json
+++ b/packages/api/test/__fixtures__/definitions/basiq.json
@@ -1,0 +1,104 @@
+{
+  "openapi": "3.0.1",
+  "info": {
+    "title": "Core",
+    "description": "All included utility endpoints for Basiq partners",
+    "version": "3.0.0"
+  },
+  "servers": [
+    {
+      "url": "https://au-api.basiq.io/"
+    }
+  ],
+  "paths": {
+    "/token": {
+      "post": {
+        "tags": ["Authentication"],
+        "summary": "Generate an auth token",
+        "description": "Use this endpoint to retrieve a token that will be passed as authorization header for Basiq API",
+        "operationId": "postToken",
+        "parameters": [
+          {
+            "name": "basiq-version",
+            "in": "header",
+            "required": true,
+            "schema": {
+              "type": "string",
+              "example": "3.0"
+            }
+          }
+        ],
+        "requestBody": {
+          "content": {
+            "application/x-www-form-urlencoded": {
+              "schema": {
+                "properties": {
+                  "scope": {
+                    "type": "string"
+                  },
+                  "userId": {
+                    "type": "string"
+                  }
+                }
+              },
+              "examples": {
+                "client_access": {
+                  "summary": "For all client side requests",
+                  "value": {
+                    "scope": "CLIENT_ACCESS",
+                    "userId": "6dd30ce4-d4ba-11ec-9d64-0242ac120002"
+                  }
+                },
+                "server_access": {
+                  "summary": "For all server side requests",
+                  "value": {
+                    "scope": "SERVER_ACCESS"
+                  }
+                }
+              }
+            }
+          }
+        },
+        "responses": {
+          "200": {
+            "description": "OK"
+          }
+        },
+        "security": [
+          {
+            "api_key": []
+          }
+        ]
+      }
+    }
+  },
+  "components": {
+    "securitySchemes": {
+      "api_key": {
+        "type": "apiKey",
+        "name": "Authorization",
+        "in": "header",
+        "x-default": "Basic NjMxMjNmMWMtZjYxMy00ZjMyLWFiYzUtYzBhZDdhYTY2YmU1OjQ3NWYwMzhkLTBlZmItNGM1ZS1iMzQ0LTAzMzYxOTkyYTRlMw=="
+      },
+      "services_token": {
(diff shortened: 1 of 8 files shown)


gratcliff (approved):

domharrington (approved):
Lgtm!

packages/httpsnippet-client-api/src/index.ts
```diff
@@ -224,10 +226,9 @@ const client: Client<APIOptions> = {
             return;
           }
         } else if (headerLower === 'accept') {
-          // If the `Accept` header here is not the default or first `Accept` header for the
-          // operations' request body then we should add it otherwise we can let the SDK handle it
-          // itself.
-          if (headers[header] === operation.getContentType()) {
+          // If the `Accept` header here is JSON-like header then we can remove it from the code
+          // snippet because `api` natively supports and prioritizes JSON over any other mime type.
+          if (matchesMimeType.json(headers[header] as string)) {
```
kanadgupta, line 231: would we also want to delete the header in the case of `headers[header] === operation.getContentType()`? i feel like it makes sense intuitively to do so?

```suggestion
          if (matchesMimeType.json(headers[header] as string) || headers[header] === operation.getContentType()) {
```

packages/api/test/__fixtures__/definitions/basiq.json
kanadgupta, line 1: not sure what pattern you generally use in this repo but i tend to prefer file names that describe the issue rather than the API it describes

kanadgupta (approved):
few minor questions below but otherwise LGTM

packages/httpsnippet-client-api/src/index.ts
erunion: This was actually a bad assumption on my part equating `Accept` to `Content-Type`. We already handle `Content-Type` headers right above this.

packages/api/test/__fixtures__/definitions/basiq.json
erunion: Generally prefer that as well but we've got a couple customer API definitions in that directory already so I think it's ok.

kanadgupta (approved):

PR #723, opened 2023-09-12
feat: deprecating the dynamic `api` entrypoint

erunion:
## 🧰 Changes

The dynamic, [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy)-based, variant of `api` has been really holding us back so this is the formal deprecation and removal of all code relating to it. It was a neat experimental hack that I am very proud of but it's time to move on to only supporting codegen here as all that work, and the TS type backings it offers, are far superior.

Rest in peace little buddy.

fixes RM-7986


packages/httpsnippet-client-api/test/index.test.ts
```diff
@@ -137,72 +128,6 @@ describe('httpsnippet-client-api', () => {
 
         expect(`${code}\n`).toStrictEqual(expected);
       });
-
-      it('should generate a functional snippet', async () => {
```
erunion, line 141: These tests used a dist of the dynamic `api` entrypoint. I don't really have any interest in trying to get this extremely gross (my fault) `vm` sandboxing work working on that so I'm deleting all this. In the future it would probably be worth fleshing out a full integration suite for all these test cases but that's all really pertinent to `api` and not code snippet generation.

packages/api/package.json
```diff
@@ -2,8 +2,6 @@
   "name": "api",
   "version": "6.1.1",
   "description": "Magical SDK generation from an OpenAPI definition 🪄",
-  "main": "./dist/index.js",
-  "types": "./dist/index.d.ts",
```
kanadgupta, line 6: :rip-cowboy:

packages/httpsnippet-client-api/test/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "noImplicitAny": false,
     "module": "Node16",
     "resolveJsonModule": true,
-    "target": "ES2022"
+    "target": "ES2015"
```
kanadgupta, line 7: what's this for?

packages/httpsnippet-client-api/test/index.test.ts
kanadgupta: would be cool to have some sort of mocking framework/library for easily testing these plugins

kanadgupta (approved):
you'll be missed, proxy variant 🫡

packages/httpsnippet-client-api/test/index.test.ts
erunion: yeah that'd be nice

PR #729, opened 2023-09-14
feat: moving node imports over to using the node: prefix

erunion:
## 🧰 Changes

This is mostly prework for the coming ESM support but this updates our Node module imports and requires to use the `node:` protocol.

I've also fixed few `any` typings in a number of spots that we missed.


packages/api/test/codegen/languages/typescript.test.ts
```diff
@@ -1,7 +1,7 @@
 import type { TSGeneratorOptions } from '../../../src/codegen/languages/typescript';
 
-import { promises as fs } from 'fs';
```
erunion, line 3: no idea why i loaded `fs/promises` like this. the things you find in small refactors...

packages/core/src/index.ts
```diff
@@ -121,7 +122,7 @@ export default class APICore {
         init.signal = controller.signal;
       }
 
-      return fetchHar(har as any, {
+      return fetchHar(har as Har, {
```
erunion, line 125: The HAR coming out of `@readme/oas-to-har` is a HAR but not a _full_ one so we need to cast this.

kanadgupta (approved):

packages/core/src/lib/parseResponse.ts
```diff
@@ -8,6 +8,7 @@ export default async function parseResponse<HTTPStatus extends number = number>(
 
   const responseBody = await response.text();
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   let data: any = responseBody;
```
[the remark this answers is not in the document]
erunion: unfort not

PR #733, opened 2023-09-20
feat: migrate `api` and `core` to ESM

kanadgupta:
| 🚥 Resolves RM-7989 |
| :------------------- |

## 🧰 Changes

This _very much (in-progress)_ PR migrates the `api` and `core` subpackages over to ESM. TKTK

## 🧬 QA & Testing

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -20,6 +20,7 @@
     "url": "https://github.com/readmeio/api.git",
     "directory": "packages/api"
   },
+  "type": "module",
   "homepage": "https://api.readme.dev",
   "bugs": {
     "url": "https://github.com/readmeio/api/issues"
@@ -48,7 +49,7 @@
     "lodash.setwith": "^4.3.2",
     "lodash.startcase": "^4.4.0",
     "make-dir": "^3.1.0",
-    "oas": "^20.4.0",
+    "oas": "^21.1.1",
     "ora": "^5.4.1",
     "prompts": "^2.4.2",
     "semver": "^7.3.8",
--- a/packages/api/test/codegen/languages/typescript.test.ts
+++ b/packages/api/test/codegen/languages/typescript.test.ts
@@ -1,4 +1,4 @@
-import type { TSGeneratorOptions } from '../../../src/codegen/languages/typescript';
+import type { TSGeneratorOptions } from '../../../src/codegen/languages/typescript.js';
 
 import fs from 'node:fs/promises';
 import path from 'node:path';
@@ -10,9 +10,9 @@ import Oas from 'oas';
 import uniqueTempDir from 'unique-temp-dir';
 import { describe, beforeEach, afterEach, it, expect, vi } from 'vitest';
 
-import TSGenerator from '../../../src/codegen/languages/typescript';
-import * as packageInfo from '../../../src/packageInfo';
-import Storage from '../../../src/storage';
+import TSGenerator from '../../../src/codegen/languages/typescript.js';
+import * as packageInfo from '../../../src/packageInfo.js';
+import Storage from '../../../src/storage.js';
 
 function assertSDKFixture(file: string, fixture: string, opts: TSGeneratorOptions = {}) {
   return async () => {
@@ -172,7 +172,7 @@ describe('typescript', () => {
       });
 
       it('should be able to make an API request (TS)', async () => {
-        const sdk = await import('../../__fixtures__/sdk/simple-ts').then(r => r.default);
+        const sdk = await import('../../__fixtures__/sdk/simple-ts/index.js').then(r => r.default);
         fetchMock.get('http://petstore.swagger.io/v2/pet/findByStatus?status=available', mockResponse.searchParams);
 
         await sdk.findPetsByStatus({ status: ['available'] }).then(({ data, status, headers, res }) => {
@@ -184,7 +184,7 @@ describe('typescript', () => {
       });
 
       it('should be able to make an API request with an `accept` header`', async () => {
-        const sdk = await import('../../__fixtures__/sdk/simple-ts').then(r => r.default);
+        const sdk = await import('../../__fixtures__/sdk/simple-ts/index.js').then(r => r.default);
         fetchMock.get('http://petstore.swagger.io/v2/pet/findByStatus?status=available', mockResponse.headers);
 
         await sdk
@@ -198,7 +198,7 @@ describe('typescript', () => {
       });
 
       it('should be able to make an API request (JS + CommonJS)', async () => {
-        const sdk = await import('../../__fixtures__/sdk/simple-js-cjs').then(r => r.default);
+        const sdk = await import('../../__fixtures__/sdk/simple-js-cjs/index.js').then(r => r.default);
         fetchMock.get('http://petstore.swagger.io/v2/pet/findByStatus?status=available', mockResponse.searchParams);
 
         await sdk.findPetsByStatus({ status: ['available'] }).then(({ data, status, headers, res }) => {
@@ -210,7 +210,7 @@ describe('typescript', () => {
       });
 
       it('should be able to make an API request (JS + ESM)', async () => {
-        const sdk = await import('../../__fixtures__/sdk/simple-js-esm').then(r => r.default);
+        const sdk = await import('../../__fixtures__/sdk/simple-js-esm/index.js').then(r => r.default);
         fetchMock.get('http://petstore.swagger.io/v2/pet/findByStatus?status=available', mockResponse.searchParams);
 
         await sdk.findPetsByStatus({ status: ['available'] }).then(({ data, status, headers, res }) => {
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -4,6 +4,7 @@
   "description": "The magic behind `api` 🧙",
   "main": "./dist/index.js",
   "types": "./dist/index.d.ts",
+  "type": "module",
   "scripts": {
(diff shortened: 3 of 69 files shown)


packages/api/package.json
```diff
@@ -2,11 +2,13 @@
   "name": "api",
   "version": "7.0.0-alpha.3",
   "description": "Magical SDK generation from an OpenAPI definition 🪄",
+  "type": "module",
```
erunion, line 5: `api` is now ESM-only. Codegen'd SDKs will support ESM and CJS but because `api` is a CLI-only tool we don't need to accommodate both module flows.

packages/api/test/codegen/languages/typescript.test.ts
```diff
@@ -197,8 +196,8 @@ describe('typescript', () => {
           });
       });
 
-      it('should be able to make an API request (JS + CommonJS)', async () => {
-        const sdk = await import('../../__fixtures__/sdk/simple-js-cjs').then(r => r.default);
+      it.skip('should be able to make an API request (JS + CommonJS)', async () => {
```
erunion, line 199: These tests that are testing CJS built SDKs no longer work because of the way that these modules are handled now. These tests will get reworked in https://github.com/readmeio/api/pull/734 when we refactor the codegen process to _properly_ build CJS exports.

packages/api/test/global.d.ts
```diff
@@ -1 +1,2 @@
-import '@api/test-utils/vitest.matchers';
+// eslint-disable-next-line import/extensions
+import '@api/test-utils/vitest.matchers.ts';
```
erunion, line 2: This looks funky but because the way the `test-utils` package is setup if we don't have this `.ts` extension here then the import for these Vitest matchers don't get properly loaded. @kanadgupta and I went through a whole pairing on session on this and settled on a method of getting it working with a very complicated `exports` object in the `package.json` for that package but that ultimately broke a lot of things downstream where we're loading definitions out of `@api/test-utils/definitions`.

Just adding this `.ts` extension, and the very minor change to the Vitest config to use `../test-utils` instead of `@api/test-utils` is a cleaner option than a flaky package config.

packages/core/package.json
```diff
@@ -2,18 +2,33 @@
...
-  "main": "./dist/index.js",
-  "types": "./dist/index.d.ts",
+  "sideEffects": false,
+  "type": "module",
+  "exports": {
+    ".": {
+      "require": "./dist/index.cjs",
+      "import": "./dist/index.js"
+    },
+    "./errors/fetchError": {
```
erunion, line 12: This is the only other export because it might be nice for folks to be able to do `instanceof FetchError` in their implementations.

packages/core/src/index.ts
```diff
@@ -35,8 +31,6 @@ type Enumerate<N extends number, Acc extends number[] = []> = Acc['length'] exte
 
 export type HTTPMethodRange<F extends number, T extends number> = Exclude<Enumerate<T>, Enumerate<F>>;
 
-export { getJSONSchemaDefaults, parseResponse, prepareAuth, prepareParams, prepareServer };
```
erunion, line 38: Not only are these utilities causing issues with the CJS export of this library but also these utilities are never used outside of core internals so they don't need to be exported.

packages/core/src/index.ts
```diff
@@ -35,8 +31,6 @@ type Enumerate<N extends number, Acc extends number[] = []> = Acc['length'] exte
...
-
 export default class APICore {
```
erunion, line 34: This default export alongside the named TS type exports is still causing issues with `attw`:

```
> @readme/api-core@7.0.0-alpha.3 attw
> attw --pack --format table-flipped


@readme/api-core v7.0.0-alpha.3

Build tools:
- typescript@^5.2.2

❓ The JavaScript appears to set both module.exports and module.exports.default for improved compatibility, but the types only reflect the latter (by using export default). This will cause TypeScript under the node16 module mode to think an extra .default property access is required, which will work at runtime but is not necessary. These types export = an object with a default property instead of using export default. https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/MissingExportEquals.md

💀 Import failed to resolve to type declarations or JavaScript files. https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/NoResolution.md


┌──────────────────────────────────────┬───────────────────────┬───────────────────────┬───────────────────┬───────────┐
│                                      │ node10                │ node16 (from CJS)     │ node16 (from ESM) │ bundler   │
├──────────────────────────────────────┼───────────────────────┼───────────────────────┼───────────────────┼───────────┤
│ "@readme/api-core"                   │ ❓ Missing `export =` │ ❓ Missing `export =` │ 🟢 (ESM)          │ 🟢        │
├──────────────────────────────────────┼───────────────────────┼───────────────────────┼───────────────────┼───────────┤
│ "@readme/api-core/errors/fetchError" │ 💀 Resolution failed  │ 🟢 (CJS)              │ 🟢 (ESM)          │ 🟢        │
├──────────────────────────────────────┼───────────────────────┼───────────────────────┼───────────────────┼───────────┤
│ "@readme/api-core/package.json"      │ 🟢 (JSON)             │ 🟢 (JSON)             │ 🟢 (JSON)         │ 🟢 (JSON) │
└──────────────────────────────────────┴───────────────────────┴───────────────────────┴───────────────────┴───────────┘
```

Everything _seems_ to work with the dist but it still worries me. I attempted to refactor the TS type exports out into a `types.ts` file, which satisfies `attw` but using that types file is another story:

```diff
diff --git a/packages/core/package.json b/packages/core/package.json
index 5d16e2b..4e95761 100644
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -13,6 +13,10 @@
       "require": "./dist/errors/fetchError.cjs",
       "import": "./dist/errors/fetchError.js"
     },
+    "./types": {
+      "require": "./dist/types.d.cts",
+      "import": "./dist/types.d.ts"
+    },
     "./package.json": "./package.json"
   },
   "main": "dist/index.cjs",
```

Screen Shot 2023-10-09 at 5 37 16 PM

packages/httpsnippet-client-api/src/index.ts
```diff
@@ -87,6 +87,7 @@ const client: Client<APIOptions> = {
     title: 'API',
     link: 'https://npm.im/api',
     description: 'Automatic SDK generation from an OpenAPI definition.',
+    extname: '.js',
```
erunion, line 90: This is a new requirement I added in `@readme/httpsnippet` when I upgraded it to support ESM.

packages/httpsnippet-client-api/test/index.test.ts
```diff
@@ -111,7 +110,7 @@ describe('httpsnippet-client-api', () => {
       it('should generate the expected snippet', async () => {
         const expected = await fs.readFile(path.join(DATASETS_DIR, snippet, 'output.js'), 'utf-8');
 
-        const code = new HTTPSnippet(mock.har).convert('node', 'api', {
+        const code = await new HTTPSnippet(mock.har).convert('node', 'api', {
```
erunion, line 113: Thanks to some native `fetch` improvements I did in `@readme/httpsnippet`, completely dropping support for funky `FormData` objects that `node-fetch@2` supports, the `convert` function there is now async.

erunion (approved):

packages/api/test/global.d.ts
kanadgupta: so weird my goodness

packages/api/tsup.config.ts
kanadgupta, line 1: hot take — do we need `tsup` at all in this subpackage if we're in a module package building ESM-only? couldn't we just rely on `tsc`?

packages/core/tsup.config.ts
kanadgupta, line 1: okay i also don't think we need to build this subpackage with `tsup` either and I think we can just rely on `tsc` — I was able to get pretty far in #734 with just the ESM build of this subpackage! i say this for several reasons:
1. to eliminate the confusing accommodations we have to make here (see [this comment](https://github.com/readmeio/api/pull/733#discussion_r1351068450))

1. since we're using `tsup` at the code-gen level, which pretty much eliminates the need for us to do any dual-export mess for downstream packages (minus packages that we need

tsconfig.base.json
kanadgupta, line 1: any reason why we no longer have `strict` mode enabled?

kanadgupta (commented):
changes look good but I don't think we need `tsup` for the `api` and `core` packages, more on this below!

packages/core/tsup.config.ts
erunion: I think? that id we dont offer cjs exports of this library then cjs codegend sdks wont be able to use it because itll be esm only and tsup wont touch it because its loaded in code as a dependency

packages/api/tsup.config.ts
erunion: Yeah i think we can just use tsc here

packages/core/tsup.config.ts
erunion: Ill change this and see how it shakes out

tsconfig.base.json
erunion: Accidental removal I think

packages/core/tsup.config.ts
erunion: slept on it for a few days and i'm a bit nervous about making `core` ESM-only with `tsc` so im going to wait until the `tsup` codegen work is done before making any refactors to the `core` build process.

PR #746, opened 2023-09-28
refactor: remove `make-dir` as a dep

kanadgupta:
## 🧰 Changes

`make-dir` isn't needed a we can call` fs.makeDirSync` instead.


erunion (approved):

kanadgupta:
@erunion the `fs` return value is slightly different from `make-dir`’s (I think one includes the full path and one is only the file name itself) and it’s not tested so we might need to do a bit of additional validation here 🫠

PR #753, opened 2023-10-06
fix(api): storage dir not being set properly

erunion:
## 🧰 Changes

Fixes a condition introduced in https://github.com/readmeio/api/pull/746 where `Storage.dir` would not get set properly when we create a new `.api/` directory.


kanadgupta (commented):
ohhh ty 🫶🏽

PR #754, opened 2023-10-11
feat(api): proper esm support and codegen overhauls

erunion:
| 🚥 Resolves RM-7990 |
| :--- |

## 🧰 Changes

* [x] Changes `api install` to no longer prompt for a language or target to install. JS is now the only export target.
* [x] Overhauls codegen internals to now use [tsup](https://npm.im/tsup) to compile our constructed TS code into a JS SDK that can be used in CJS and ESM environments.
  * We were previously handling this with a lot of very touchy `ts-morph` hacks that didn't _really_ support ESM environments.
* [x] Refactored codegen to now store generated SDKs within a `src/` directory. 
  * Doing this mostly for aesthetic reasons now that we're going to have a `dist/` directory post-`tsup` compilation.
* [x] Upgrades `ts-morph` to the latest release. https://github.com/readmeio/api/pull/752 

## 🧬 QA & Testing

* [ ] Can you create an SDK?
* [ ] Can you load the SDK from a `.cjs`, `.mjs`, `.js`, and `.ts` example file?

PR #756, opened 2023-10-12
fix(api): nestling generated schemas into a new `schemas/` directory

erunion:
## 🧰 Changes

With the ESM and [tsup](https://npm.im/tsup) work we're doing in https://github.com/readmeio/api/pull/754 I discovered that the way we're currently generating `schemas.ts` might result in an uncompilable file due to the way we're processing OpenAPI component schemas top-down.

Basically what happens is that if you use a `$ref` component in a schema and that component exists further down in the API definition we'll generate a `schemas.ts` in the order they were specified resulting in us attempting to use `const` that has not been created yet:

screen_shot_2023-10-11_at_5 16 20_pm_720

There are a number of ways that we can fix this but @kanadgupta had the great idea to shove these schemas into individual files in a new `schemas/` directory. The result is that codegen'd SDKs have _n_ more files now but they'll at least be able to be compiled with tooling like `tsup`.


packages/api/test/__fixtures__/sdk/simple-js-cjs/schemas.d.ts
```diff
@@ -1,196 +1,5 @@
...
-            readonly maximum: 9223372036854776000;
-        };
-        readonly name: {
-            readonly type: "string";
-        };
-    };
-    readonly title: "Tag";
-    readonly "x-readme-ref-name": "Tag";
-};
+import Category from './schemas/Category';
```
erunion, line 1: This is wrong because CJS and ESM compilation with `ts-morph` don't create this `schemas/` directory but because we're overhauling the CJS and ESM compilation work with https://github.com/readmeio/api/pull/754 I'm going to leave this breakage alone.

kanadgupta:
hell yeah 🫶🏽

PR #759, opened 2023-10-16
fix: minor `tsup` codegen fixes

kanadgupta:
## 🧰 Changes

Was playing around in [the sandbox](https://github.com/readmeio/tmp-api-tsup-sandbox) and had to make a few minor tweaks to get it up and running properly.

## 🧬 QA & Testing

Does the sandbox work?

### DIFF
--- a/packages/api/bin/api.js
+++ b/packages/api/bin/api.js
@@ -1,2 +1,2 @@
-#!/usr/bin/env node
+#!/usr/bin/env NODE_OPTIONS=--no-warnings node
 import '../dist/bin.js';
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -38,6 +38,7 @@
     "swagger"
   ],
   "dependencies": {
+    "@readme/api-core": "file:../core",
     "@readme/openapi-parser": "^2.4.0",
     "chalk": "^4.1.2",
     "commander": "^11.0.0",
@@ -60,7 +61,6 @@
   },
   "devDependencies": {
     "@api/test-utils": "file:../test-utils",
-    "@readme/api-core": "file:../core",
     "@readme/oas-examples": "^5.12.0",
     "@types/js-yaml": "^4.0.5",
     "@types/lodash.camelcase": "^4.3.7",
--- a/packages/api/src/codegen/languages/typescript/index.ts
+++ b/packages/api/src/codegen/languages/typescript/index.ts
@@ -17,6 +17,7 @@ import type { JsonObject, PackageJson, TsConfigJson } from 'type-fest';
 
 import path from 'node:path';
 
+import corePkg from '@readme/api-core/package.json' assert { type: 'json' };
 import execa from 'execa';
 import setWith from 'lodash.setwith';
 import semver from 'semver';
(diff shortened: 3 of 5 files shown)


packages/api/tsconfig.json
```diff
@@ -1,9 +1,10 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "esModuleInterop": true,
     "lib": ["DOM", "DOM.Iterable", "ES2023"],
-    "outDir": "dist/"
+    "module": "NodeNext",
```
kanadgupta, line 5: apparently JSON module import assertions don't support `module: Node16`. annoying!

packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -167,6 +168,10 @@ export default class TSGenerator extends CodeGenerator {
   async compile(storage: Storage, opts: InstallerOptions = {}): Promise<void> {
     const installDir = storage.getIdentifierStorageDir();
 
+    await execa('npm', ['install', 'tsup', 'typescript', '-D'], {
```
kanadgupta, line 171: we need TS as a dev-dep so the `tsup` build runs properly:

CleanShot 2023-10-16 at 15 09 46@2x

packages/api/package.json
```diff
@@ -38,6 +38,7 @@
     "swagger"
   ],
   "dependencies": {
+    "@readme/api-core": "file:../core",
```
kanadgupta, line 41: we grab the package version from `@readme/api-core`, which is why it's now a regular dep (but will defer to you on the best way to bring this in)

packages/api/tsconfig.json
```diff
@@ -1,9 +1,10 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "esModuleInterop": true,
```
kanadgupta, line 4: this is unnecessary, already enabled in the base config

packages/api/tsconfig.json
```diff
@@ -1,9 +1,10 @@
...
     "lib": ["DOM", "DOM.Iterable", "ES2023"],
-    "outDir": "dist/"
+    "module": "NodeNext",
+    "outDir": "dist/",
+    "resolveJsonModule": true
```
kanadgupta, line 7: needed for the JSON import

packages/api/bin/api.js
```diff
@@ -1,2 +1,2 @@
-#!/usr/bin/env node
+#!/usr/bin/env NODE_OPTIONS=--no-warnings node
```
kanadgupta, line 1: needed to silence this warning now that we're importing JSON modules

```
(node:16233) ExperimentalWarning: Import assertions are not a stable feature of the JavaScript language. Avoid relying on their current behavior and syntax as those might change in a future version of Node.js.
(Use `node --trace-warnings ...` to show where the warning was created)
(node:16233) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
```

packages/api/src/codegen/languages/typescript/index.ts
kanadgupta: we probably want to include this in that list of packages along with some CLI explainers as to why we need said packages, but this PR is just intended to get it working properly

erunion (approved):

PR #760, opened 2023-10-16
refactor(core): removing `oas` as a dependency on codegen'd SDKs

erunion:
## 🧰 Changes

Because `oas` is only used in codegen'd SDKs to do `Oas.init()` and supply an instance of `oas` to `@readme/api-core` we can lessen some steps, and a hard dependency on `oas` here by refactoring `@readme/api-core` to instead accept an API definition as its first argument.

### DIFF
--- a/packages/api/src/codegen/languages/typescript/index.ts
+++ b/packages/api/src/codegen/languages/typescript/index.ts
@@ -99,11 +99,6 @@ export default class TSGenerator extends CodeGenerator {
         url: 'https://npm.im/json-schema-to-ts',
         version: '^2.9.2',
       },
-      oas: {
-        reason: 'Used within `@readme/api-core` and is also loaded for TypeScript types.',
-        url: 'https://npm.im/oas',
-        version: '^23.0.0',
-      },
     };
 
     this.project = new Project({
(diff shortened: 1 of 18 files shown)


packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -265,24 +260,19 @@ export default class TSGenerator extends CodeGenerator {
         defaultImport: 'type { ConfigOptions, FetchResponse, HTTPMethodRange }',
         moduleSpecifier: '@readme/api-core',
       },
-      { defaultImport: 'Oas', moduleSpecifier: 'oas' },
```
erunion, line 268: This is the `import Oas from 'oas'` line in SDKs.

packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -265,24 +260,19 @@ export default class TSGenerator extends CodeGenerator {
...
     ]);
 
     // @todo add TOS, License, info.* to a docblock at the top of the SDK.
     this.sdk = sourceFile.addClass({
       name: 'SDK',
-      properties: [
-        { name: 'spec', type: 'Oas' },
-        { name: 'core', type: 'APICore' },
-      ],
+      properties: [{ name: 'core', type: 'APICore' }],
```
erunion, line 270: This creates the `spec: Oas` and `core: APICore` properties in the main SDK class.

kanadgupta (approved):
ohhhh this is so much better 🫶🏽

PR #761, opened 2023-10-16
feat(codegen): required devDeps are now listed in `package.json`

erunion:
## 🧰 Changes

Per feedback in https://github.com/readmeio/api/pull/754 this slightly refactors the `package.json` file we generate for SDKs to now list the dev-only dependencies that are required to build an SDK.

I've also added in a `prepare` script to these for rebuilding the compiled SDK, as well as a `files` array for when/if a codegen'd SDK is published to NPM.

### DIFF
--- a/packages/api/src/codegen/codegenerator.ts
+++ b/packages/api/src/codegen/codegenerator.ts
@@ -28,8 +28,9 @@ export default abstract class CodeGenerator {
   requiredPackages!: Record<
     string,
     {
+      dependencyType: 'production' | 'development';
       reason: string;
-      url: string;
+      url?: string;
       version: string;
     }
   >;
--- a/packages/api/src/codegen/languages/typescript/index.ts
+++ b/packages/api/src/codegen/languages/typescript/index.ts
@@ -83,6 +83,7 @@ export default class TSGenerator extends CodeGenerator {
 
     this.requiredPackages = {
       '@readme/api-core': {
+        dependencyType: 'production',
         reason: "The core magic of your codegen'd SDK and is what is used for making requests.",
         url: 'https://npm.im/api',
         version:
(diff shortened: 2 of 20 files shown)


packages/api/src/codegen/codegenerator.ts
```diff
@@ -28,8 +28,9 @@ export default abstract class CodeGenerator {
   requiredPackages!: Record<
     string,
     {
+      dependencyType: 'production' | 'development';
       reason: string;
-      url: string;
+      url?: string;
```
erunion, line 33: Made `url` optional because it feels silly to write out https://typescriptlang.org/ as the URL for TS.

packages/api/src/codegen/codegenerator.ts
```diff
@@ -28,8 +28,9 @@ export default abstract class CodeGenerator {
   requiredPackages!: Record<
     string,
     {
+      dependencyType: 'production' | 'development';
```
erunion, line 31: Made this as `dependencyType` instead of `devDependency: boolean` so it's not specific to NPM.

kanadgupta: yeah that's a great call

packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -95,10 +96,22 @@ export default class TSGenerator extends CodeGenerator {
...
+      tsup: {
+        dependencyType: 'development',
+        reason: "Used for compiling your codegen'd SDK into code that can be used in JS environments.",
+        url: 'https://tsup.egoist.dev/',
+        version: '^7.2.0',
+      },
+      typescript: {
+        dependencyType: 'development',
+        reason: 'Required for `tsup`.',
+        version: '^5.2.2',
```
kanadgupta, line 113: if and when we improve the codegen for our test fixtures (either via a build script or something), it'd be cool if we can dynamically pull in the versions from our dependencies

kanadgupta (approved):

packages/api/src/codegen/languages/typescript/index.ts
erunion: 100%

PR #763, opened 2023-10-17
feat(cli): prompting the user to use `info.title` for their installed SDK

### LINKED ISSUE #676: Prefill the identifier with `api.info.title`
Right now the CLI asks you what you want to identify your API SDK as and we give `petstore` as an example, it would be nice if we prefilled that with a slugified + npm-friendly version of whatever is in `info.title`.

erunion:
| 🚥 Resolves #676 |
| :------------------- |

## 🧰 Changes

This slightly changes the installation process to now prompt the user if they would like to install their SDK using the `info.title` value from the API definition they supplied:

Screen Shot 2023-10-16 at 5 33 24 PM

Note that this work only applies if they supply a non-ReadMe API Registry value to `npx api install`:

```bash
npx api install --identifier buster demo.json # will continue to use "buster"
npx api install @developers/v2.0#17273l2glm9fq4l # will continue to use "developers"
npx api install demo.json # will prompt the user if they want to use `info.title`
```


packages/api/src/commands/install.ts
```diff
@@ -51,47 +71,29 @@ cmd
...
+      identifier = uslug(oas.api.info.title, { lower: true });
+
+      // let useInfoAsIdentifier;
+      const { value: confirmation } = await confirmProposedIdentifier(oas.api.info.title, identifier);
+      if (!confirmation) {
+        // If they don't like what we picked from the spec info doc then let's have them tell us
+        // what they want.
+        ({ value: identifier } = await promptForIdentifier(false));
+      }
     } else {
```
erunion, line 84: We should never get to this `else` case because `info.title` is required in an OpenAPI definition but just in case we don't have it for whatever reason, we'll just fallback to asking them what they want -- though codegen will likely fail immediately after this.

erunion:
Revised this a bit and now there's just a single identifier question with a new initial value based on `info.title` if it's present.

Screen Shot 2023-10-17 at 9 59 30 AM

packages/api/src/commands/install.ts
```diff
@@ -43,6 +44,27 @@ cmd
       // logger(`It looks like you already have this API installed. Would you like to update it?`);
     }
 
+    let spinner = ora('Fetching your API').start();
```
kanadgupta, line 47:
```suggestion
    let spinner = ora('Fetching your API definition').start();
```

kanadgupta (approved):

PR #764, opened 2023-10-17
feat(core): moving `json-schema-to-ts` directly into the `core` package

erunion:
| 🚥 Resolves RM-8176 |
| :------------------- |

## 🧰 Changes

A slight refactor to move the `json-schema-to-ts` dependency in codegen'd SDKs into `@readme/api-core`.


packages/core/src/lib/index.ts
kanadgupta, line 1: is this barrel file going to be an issue from a tree-shaking standpoint?

kanadgupta (approved):

packages/core/src/lib/index.ts
erunion: everything in this file is used in `@readme/api-core` so i don't think anything would be treeshaken out.

PR #765, opened 2023-10-17
chore(deps): bumping all out of date dependencies

erunion:
| 🚥 Resolves RM-8175 |
| :------------------- |

## 🧰 Changes

Now that we can handle ESM libraries here we're finally able to update a grip of dependencies that are terribly out of date.


packages/core/src/lib/prepareParams.ts
```diff
@@ -111,7 +111,7 @@ function processFile(
       });
     });
   } else if (file instanceof stream.Readable) {
-    return getStream.buffer(file).then(buffer => {
+    return getStreamAsBuffer(file).then(buffer => {
```
kanadgupta, line 114: oooh better tree-shaking here? maybe?

.github/dependabot.yml
kanadgupta, line 1: 😌

kanadgupta (approved):

packages/core/src/lib/prepareParams.ts
erunion: Yea likely

PR #767, opened 2023-10-18
feat(httpsnippet-client-api): updates for v7 compatibility + custom variable support

erunion:
| 🚥 Resolves RM-3674 |
| :------------------- |

## 🧰 Changes

* [x] Updates snippet generation for v7 compatibility where we no longer have the dynamic `require('api')('...')` accessor.
* [x] Adds support for supplying snippets a custom variable (folks apparently don't like us assigning the retrieved SDK to `sdk`).
* [x] Changes the snippet syntax from `require` over to `import` because 2023, and also codegen now supports CJS and ESM.

### DIFF
--- a/packages/api/src/fetcher.ts
+++ b/packages/api/src/fetcher.ts
@@ -10,6 +10,8 @@ export default class Fetcher {
   uri: string | OASDocument;
 
   /**
+   * @note This regex also exists in `httpsnippet-client-api`.
+   *
    * @example @petstore/v1.0#n6kvf10vakpemvplx
    * @example @petstore#n6kvf10vakpemvplx
    */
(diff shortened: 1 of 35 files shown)


kanadgupta (commented):
i don't see any changes related to this installation step:

CleanShot 2023-10-18 at 13 55 11@2x

does that happen in another repo?

erunion:
@kanadgupta Yeah that's part of `@readme/oas-to-snipppet`. https://github.com/readmeio/oas-to-snippet/blob/main/src/supportedLanguages.ts#L149

I still need to update that library to support dual exports and will update the install step then.

kanadgupta (approved):

PR #768, opened 2023-10-18
fix(httpsnippet-client-api): retain casing for non-reserved headers

erunion:
| 🚥 Resolves RM-8153 |
| :------------------- |

## 🧰 Changes

This fixes a quirk in `httpsnippet-client-api` where even though HTTP headers are case-insensitive the TS types we generate for `api` are not and we are currently lowercasing all non-reserved header names we add into code snippets.

image

### DIFF
--- a/packages/httpsnippet-client-api/test/__datasets__/headers/index.ts
+++ b/packages/httpsnippet-client-api/test/__datasets__/headers/index.ts
@@ -17,8 +17,12 @@ const mock: SnippetMock = {
         value: 'Bar',
       },
       {
-        name: 'X-Bar',
+        name: 'x-bar',
         value: 'foo',
+      },
+      {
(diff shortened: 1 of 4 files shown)


packages/httpsnippet-client-api/test/__datasets__/headers/index.ts
```diff
@@ -17,9 +17,13 @@ const mock: SnippetMock = {
         value: 'Bar',
       },
       {
-        name: 'X-Bar',
```
erunion, line 20: Changed this casing because `X-Bar` doesn't exist in the spec for this dataset and how we construct HAR objects in the API Explorer we won't ever translate `x-bar` to `X-Bar` (because `X-Bar` isn't in the spec).

PR #769, opened 2023-10-19
feat(api): codegen'd sdks now look a little bit nicer

erunion:
## 🧰 Changes

This is a big changeset but the only changes that happened here are the following:

* [x] Updated the TS codegen test suite to allow us to ad-hoc regenerate failing SDK fixtures by running vitest with a `UPDATE_FIXTURES=1` environment variable.
* [x] Fixed a quirk where in a couple cases a line-terminating semicolon would be annoyingly be placed on the following line.
* [x] Generated JSON Schemas now look a lot nicer thanks to `ts-morph`'s formatting API.[^1]
  * I have not noticed any adverse impacts to codegen performance by enabling this.

[^1]: https://ts-morph.com/manipulation/formatting


packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -230,20 +230,19 @@ export default class TSGenerator extends CodeGenerator {
...
-
-      // Because we're returning the raw source files for TS generation we also need to separately
-      // emit out our declaration files so we can put those into a separate file in the installed
-      // SDK directory.
-      ...this.project
-        .emitToMemory({ emitOnlyDtsFiles: true })
-        .getFiles()
-        .map(sourceFile => ({
-          [path.basename(sourceFile.filePath)]: sourceFile.text,
-        })),
```
erunion, line 246: This is no longer used and is safe to be removed. It was a remaining artifact of the previous CJS and (attempted) ESM export work where we were generating `.d.ts` files ourselves.

packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -383,7 +382,7 @@ sdk.server('https://eu.api.example.com/v14');`),
           name: 'createSDK',
           initializer: writer => {
             // `ts-morph` doesn't have any way to cleanly create an IIFE.
-            writer.writeLine('(() => { return new SDK(); })()');
+            writer.write('(() => { return new SDK(); })()');
```
erunion, line 385: Changing this from `.writeLine()` to `.write()` fixes a quirk where the ending `;` would get placed on the following line because `.writeLine()` terminates the thing we're writing with a `\n`.

packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -549,7 +550,7 @@ sdk.server('https://eu.api.example.com/v14');`),
               // need to clean them up.
               str = str.replace(REF_PLACEHOLDER_REGEX, '$1');
 
-              writer.writeLine(`${str} as const`);
+              writer.write(`${str} as const`);
```
erunion, line 553: Same as the other semicolon fix. These schemas were getting generated like the following:

```ts
const schema = {"type": "string"} as const
;
```

They're now this:

```ts
const schema = {"type": "string"} as const;
```

<sub>its the little things</sub>

packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -230,20 +230,19 @@ export default class TSGenerator extends CodeGenerator {
         let filePath = sourceFile.getFilePath().toString();
         filePath = filePath.substring(1);
 
+        /**
+         * It's not Prettier-level of nice but  `ts-morph` offers a method of using the TS
```
kanadgupta, line 234:
```suggestion
         * It's not Prettier-level of nice but `ts-morph` offers a method of using the TS
```

packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -230,20 +230,19 @@ export default class TSGenerator extends CodeGenerator {
...
+         * formatter for formatting code which we can use to make our generated SDK not look like
+         * total garbage.
+         *
+         * @see {@link https://ts-morph.com/manipulation/formatting}
+         */
+        sourceFile.formatText();
```
kanadgupta, line 240: nice, TIL! kinda curious about the performance of this vs prettier vs others (but don't think we need to investigate that any time soon)

kanadgupta (approved):

PR #773, opened 2023-10-19
feat(api): adding a `$schema` property to generated `api.json` lockfiles

erunion:
| 🚥 Resolves RM-8186 |
| :------------------- |

## 🧰 Changes

This adds a new `$schema` property to the `api.json` lockfiles we generate when installing an API.

## 🧬 QA & Testing

Here it is validating:

Screen Shot 2023-10-19 at 11 54 17 AM

And what it looks like after instaling an API[^1]:

Screen Shot 2023-10-19 at 12 04 11 PM

### DIFF
--- a/packages/api/src/storage.ts
+++ b/packages/api/src/storage.ts
@@ -3,6 +3,7 @@ import type { OASDocument } from 'oas/rmoas.types';
 import fs from 'node:fs';
 import path from 'node:path';
 
+import semver from 'semver';
 import ssri from 'ssri';
 import validateNPMPackageName from 'validate-npm-package-name';
 
@@ -79,7 +80,10 @@ export default class Storage {
   }
 
   static getDefaultLockfile(): Lockfile {
+    const majorVersion = semver.parse(PACKAGE_VERSION)?.major || 'latest';
+
     return {
+      $schema: `https://unpkg.com/api@${majorVersion}/schema.json`,
       version: '1.0',
       apis: [],
     };
@@ -288,7 +292,15 @@ export default class Storage {
   }
 }
 
+/**
+ * @see schema.json
+ */
 interface Lockfile {
+  $schema: string;
+
+  /**
+   * The list of installed APIs.
+   */
   apis: LockfileAPI[];
 
   /**
@@ -298,9 +310,12 @@ interface Lockfile {
   version: '1.0';
 }
 
+/**
+ * @see schema.json
+ */
 interface LockfileAPI {
   /**
-   * A unique identifier of the API. This'll be used to do requires on `@api/<identifier>` and also
+   * A unique identifier of the API. This'll be used to do imports on `@api/<identifier>` and also
    * where the SDK code will be located in `.api/apis/<identifier>`.
    *
    * @example petstore
--- a/packages/api/test/storage.test.ts
+++ b/packages/api/test/storage.test.ts
@@ -5,10 +5,12 @@ import fs from 'node:fs/promises';
 import path from 'node:path';
 
 import { loadSpec } from '@api/test-utils';
+import Ajv from 'ajv';
 import fetchMock from 'fetch-mock';
 import uniqueTempDir from 'unique-temp-dir';
 import { describe, beforeAll, beforeEach, afterEach, it, expect } from 'vitest';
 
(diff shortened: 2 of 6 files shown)


packages/api/src/storage.ts
```diff
@@ -79,7 +80,10 @@ export default class Storage {
   }
 
   static getDefaultLockfile(): Lockfile {
+    const majorVersion = semver.parse(PACKAGE_VERSION)?.major || 'latest';
+
     return {
+      $schema: `https://unpkg.com/api@${majorVersion}/schema.json`,
```
erunion, line 86: I'm using a major version here because I don't really want the schema URL to change with every minor and patch release we push out. However using a major here also means that until we tag a v7 this URL isn't going to work as `api@7` doesn't match `api@7.0.0-beta.0`.

For example see that https://unpkg.com/api@7.0.0-beta.0/package.json works but https://unpkg.com/api@7/package.json does not.

I think I'm okay with this? I don't really want to have code in here to detect if `PACKAGE_VERSION` is a beta or alpha and use the full version.

packages/api/src/storage.ts
```diff
@@ -288,7 +292,15 @@ export default class Storage {
   }
 }
 
+/**
+ * @see schema.json
+ */
 interface Lockfile {
```
kanadgupta, line 298: maybe we just use `json-schema-to-ts` to generate this interface? 👀

packages/api/src/storage.ts
kanadgupta: yeah agreed, esp since we're capturing exact versions in the `installerVersion` value!

kanadgupta (commented):
love this 🫶🏽 thoughts on having AJV validate the generated output against our schema?

packages/api/src/storage.ts
kanadgupta: actually wait — since we technically version our `api.json` independently of our `api` package, shouldn't it point to something like `https://unpkg.com/api@schema-v1/package.json`? thoughts on maintaining a `dist-tag` for that?

packages/api/src/storage.ts
erunion: `json-schema-to-ts` doesn't seem to like it for some reason. might be a good thing to refactor at some point.

Screen Shot 2023-10-19 at 12 20 48 PM

packages/api/src/storage.ts
erunion: its technically versioned independently of `api` but we'd only ever ship changes to it with an `api` release so they're kind of tied to each other. i feel like unless a `schema-v1` dist-tag process were built into our lerna config we'd quickly forget to update it.

erunion: or it could be released as its own package but that feels like a lot of overhead for a json file

packages/api/src/storage.ts
erunion: talked offline and supplying a `.json` file to `json-schema-to-ts` is a TS limitation. https://github.com/ThomasAribart/json-schema-to-ts/blob/main/documentation/FAQs/does-json-schema-to-ts-work-on-json-file-schemas.md

packages/api/src/storage.ts
kanadgupta: talked offline and we'll stick with the `api` major version tags as our reference!

packages/api/test/storage.test.ts
```diff
@@ -100,6 +102,28 @@ describe('storage', () => {
     });
   });
 
+  describe('#getLockfile', () => {
+    it('should retrieve a lockfiles that match our schema', async () => {
+      fetchMock.get('https://dash.readme.com/api/v1/api-registry/n6kvf10vakpemvplx', petstoreSimple);
+
+      // `ajv` has funky types in ESM environments. https://github.com/ajv-validator/ajv/issues/2047
+      // eslint-disable-next-line new-cap
+      const ajv = new Ajv.default();
```
erunion, line 111: Can fix this by setting `moduleResolution` in `test/tsconfig.json` to `Bundler` but because our `module` in the root is `NodeNext` they then become incompatible and TS throws errors. I'm fine leaving this `.default()` quirk in here.

kanadgupta (approved):

PR #775, opened 2023-10-19
feat(api): creation of a new `list` command

erunion:
| 🚥 Resolves RM-4124 |
| :------------------- |

## 🧰 Changes

This introduces a new `list` command for listing out any API SDKs that you may have installed.

```
Usage: api list|ls [options]

list any installed API SDKs

Options:
  -h, --help  display help for command
```

Additionally I've also added a couple new things into the lockfile:

* [x] `private`: We'll use this later in an NPM publishing flow but this'll tell us (and the user) that when they ran `npx api install <uri>` they chose to not publish it and only install it as a private package.
* [x] `createdAt`: ISO date for everyone to know when an API SDK was originally generated + installed.
* [x] `updatedAt`: Currently unused but this will be factored into the coming `rebuild` command and will keep track of when an API SDK was regenerated or otherwise updated.

We're also removing `version` from the lockfile in favor of the recently added `$schema` property. Having two versioned fields in the lockfile felt like too much complexity.

Here's what it all looks like:

Screen Shot 2023-10-19 at 1 55 17 PM

And if you don't have any APIs installed:

Screen Shot 2023-10-19 at 1 55 27 PM

### DIFF
--- a/packages/api/schema.json
+++ b/packages/api/schema.json
@@ -3,18 +3,18 @@
   "title": "api storage lockfile",
   "description": "See https://api.readme.dev/docs",
   "type": "object",
-  "required": ["apis", "version"],
+  "required": ["apis"],
+  "additionalProperties": false,
   "properties": {
+    "$schema": {
+      "type": "string"
+    },
     "apis": {
       "type": "array",
       "description": "The list of installed APIs",
       "items": {
         "$ref": "#/definitions/api"
       }
-    },
-    "version": {
-      "type": "string",
-      "description": "The current `api.json` schema version."
     }
   },
   "definitions": {
--- a/packages/api/schema.json
+++ b/packages/api/schema.json
@@ -20,8 +20,14 @@
   "definitions": {
     "api": {
       "type": "object",
-      "required": ["identifier", "installerVersion", "integrity"],
+      "required": ["createdAt", "identifier", "installerVersion", "integrity"],
       "properties": {
+        "createdAt": {
+          "type": "string",
+          "format": "date-time",
+          "description": "The date that this SDK was installed.",
+          "examples": ["2023-10-19T20:35:39.268Z"]
+        },
         "identifier": {
           "type": "string",
           "description": "A unique identifier of the API. This'll be used to do imports on `@api/<identifier>` and also where the SDK code will be located in `.api/apis/<identifier>`",
@@ -39,6 +45,10 @@
             "sha512-ld+djZk8uRWmzXC+JYla1PTBScg0NjP/8x9vOOKRW+DuJ3NNMRjrpfbY7T77Jgnc87dZZsU49robbQfYe3ukug=="
           ]
         },
+        "private": {
+          "type": "boolean",
+          "description": "Was this SDK installed as a private, unpublished, package to the filesystem?"
+        },
         "source": {
           "type": "string",
           "description": "The original source that was used to generate the SDK with.",
@@ -47,6 +57,12 @@
             "./petstore.json",
             "@developers/v2.0#nysezql0wwo236"
           ]
+        },
+        "updatedAt": {
+          "type": "string",
+          "format": "date-time",
+          "description": "The date that this SDK was last rebuilt or updated.",
+          "examples": ["2023-10-19T20:35:39.268Z"]
         }
       }
     }
--- a/packages/api/src/commands/list.ts
+++ b/packages/api/src/commands/list.ts
@@ -32,7 +32,7 @@ cmd
 
       logger(`source: ${chalk.grey(api.source)}`);
       logger(`installer version: ${chalk.grey(api.installerVersion)}`);
-      logger(`created at: ${chalk.grey(api.createdAt)}`);
+      logger(`created at: ${chalk.grey(api.createdAt || 'n/a')}`);
       if (api.updatedAt) {
         logger(`updated at: ${chalk.grey(api.updatedAt)}`);
       }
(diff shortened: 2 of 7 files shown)


packages/api/src/commands/list.ts
erunion, line 1: We don't have any way to test CLI commands right now so I didn't add any tests for this.

packages/api/test/storage.test.ts
```diff
@@ -6,6 +6,7 @@ import path from 'node:path';
 
 import { loadSpec } from '@api/test-utils';
 import Ajv from 'ajv';
+import addFormats from 'ajv-formats';
```
erunion, line 9: AJV by default doesn't support `format: date-time`. 🚯

packages/api/test/storage.test.ts
```diff
@@ -109,6 +110,8 @@ describe('storage', () => {
       // `ajv` has funky types in ESM environments. https://github.com/ajv-validator/ajv/issues/2047
       // eslint-disable-next-line new-cap
       const ajv = new Ajv.default();
+      addFormats.default(ajv);
```
erunion, line 113: AJV by default doesn't support `format: date-time`. 🚯

packages/api/src/commands/list.ts
```diff
@@ -0,0 +1,42 @@
...
+import Storage from '../storage.js';
+
+const cmd = new Command();
+cmd
+  .name('list')
+  .alias('ls')
+  .description('list any installed API SDKs')
+  .action(async () => {
+    // We need to preload the storage system so we can grab the lockfile.
+    Storage.setStorageDir();
```
erunion, line 14: This will end up creating the `.api/` directory, and an empty `api.json` file, if it doesn't exist which is probably fine? If you're using `npx api ls` to see if you have APIs installed then you're likely already pretty familiar with the library and having a sort of empty `.api/` directory might not be a big deal.

packages/api/src/commands/list.ts
```diff
@@ -0,0 +1,42 @@
+import chalk from 'chalk';
+import { Command } from 'commander';
+
+import logger from '../logger.js';
+import Storage from '../storage.js';
+
+const cmd = new Command();
+cmd
+  .name('list')
```
erunion, line 9: When thinking about this command last year I originally had plans to support `npx api list petstore` but don't really think that's worth it anymore. This list of installed APIs isn't ever going to be the size of something like `npm ls` so filtering isn't something I can ever see anybody ever want.

packages/api/schema.json
```diff
@@ -20,8 +20,14 @@
   "definitions": {
     "api": {
       "type": "object",
-      "required": ["identifier", "installerVersion", "integrity"],
+      "required": ["createdAt", "identifier", "installerVersion", "integrity"],
```
kanadgupta, line 23: this would make this a breaking change, right?

packages/api/src/commands/list.ts
```diff
@@ -0,0 +1,42 @@
...
+  .alias('ls')
+  .description('list any installed API SDKs')
+  .action(async () => {
+    // We need to preload the storage system so we can grab the lockfile.
+    Storage.setStorageDir();
+
+    const lockfile = Storage.getLockfile();
+
+    if (!lockfile.apis.length) {
+      logger('😔 You do not have any API SDKs installed.');
```
kanadgupta, line 19:
```suggestion
      logger('😔 You do not have any SDKs installed.');
```

packages/api/src/storage.ts
```diff
@@ -314,6 +316,13 @@ interface Lockfile {
  * @see schema.json
  */
 interface LockfileAPI {
+  /**
+   * The date that this SDK was installed.
```
kanadgupta, line 320: i continue to be so sad that `json-schema-to-ts` didn't work out lol

packages/api/test/storage.test.ts
kanadgupta: wooof

packages/api/src/commands/list.ts
```diff
@@ -0,0 +1,42 @@
...
+      }
+
+      logger(chalk.yellow.underline(api.identifier));
+      if (api.private) {
+        logger(`package name (${chalk.red('private')}): ${chalk.grey(`@api/${api.identifier}`)}`);
+      }
+
+      logger(`source: ${chalk.grey(api.source)}`);
+      logger(`installer version: ${chalk.grey(api.installerVersion)}`);
+      logger(`created at: ${chalk.grey(api.createdAt)}`);
```
kanadgupta, line 35: since this technically might be undefined for older users (or maybe we wrap this in an if-statement similar to what you have for `updatedAt`:

```suggestion
      logger(`created at: ${chalk.grey(api.createdAt || 'n/a')}`);
```

packages/api/src/commands/list.ts
```diff
@@ -0,0 +1,42 @@
...
+      if (api.updatedAt) {
+        logger(`updated at: ${chalk.grey(api.createdAt)}`);
```
kanadgupta, line 37:
```suggestion
        logger(`updated at: ${chalk.grey(api.updatedAt)}`);
```

kanadgupta (changes requested):
LGTM for the most part — mostly small changes!

packages/api/schema.json
erunion: hm yeah good point. i'll bump the lockfile to v2 and for lockfile v1 users if it's missing i'll not print anything.

erunion: talked offline and we're dropping `version` instead from the lockfile and are relying just on the versioned `$schema`.

packages/api/src/storage.ts
erunion: i'm already to the point of thinking of building out some script like the `src/packageInfo.ts` build process that creates `schema.json` when we build dists

kanadgupta: YES

kanadgupta (approved):
LGTM!

PR #776, opened 2023-10-20
feat(api): addition of a new uninstall command

erunion:
| 🚥 Resolves RM-4127 |
| :------------------- |

## 🧰 Changes

This creates a new `uninstall` command that folks can use to uninstall codegen'd SDKs from their codebase. Not only will it run `npm uninstall @api/<identifier>` it will also delete the generated SDK out of the `.api/` directory. Here's what it all looks like.

Screen Shot 2023-10-19 at 11 28 26 PM

Screen Shot 2023-10-19 at 11 30 25 PM

Screen Shot 2023-10-19 at 11 29 33 PM


packages/api/schema.json
```diff
@@ -45,6 +45,12 @@
             "sha512-ld+djZk8uRWmzXC+JYla1PTBScg0NjP/8x9vOOKRW+DuJ3NNMRjrpfbY7T77Jgnc87dZZsU49robbQfYe3ukug=="
           ]
         },
+        "language": {
```
erunion, line 48: Keeping track of what language we generated the SDK for will allow us to know how to uninstall it from that languages package manager. And because this didn't exist before this we'll default to `js` everywhere we access this.

packages/api/src/commands/list.ts
```diff
@@ -30,6 +30,7 @@ cmd
         logger(`package name (${chalk.red('private')}): ${chalk.grey(`@api/${api.identifier}`)}`);
       }
 
+      logger(`language: ${chalk.grey(api.language || SupportedLanguages.JS)}`);
```
erunion, line 33: Screen Shot 2023-10-19 at 11 35 45 PM

packages/api/src/storage.ts
```diff
@@ -226,30 +282,17 @@ export default class Storage {
   }
 
   /**
-   * @example <caption>Storage directory structure</caption>
```
erunion, line 229: Removed this because I don't want to have to maintain it alongside the same tree in `docs/how-it-works.md`.

docs/how-it-works.md
```diff
@@ -22,23 +22,28 @@ The `.api/` directory is where the CLI installation process stores all of its in
...
 ```
 .api/
-├── api.json             // The `package-lock.json` equivalent that records
-|                        // everything that's installed, when it was installed,
-|                        // what the original source was, and what version of
-|                        // `api` was used.
+├── api.json               // The `package-lock.json` equivalent that records
+|                          // everything that's installed, when it was installed,
+|                          // what language was generated, what the original
+|                          // source was, and what version of `api` was used.
```
kanadgupta, line 28: just b/c i don't love the idea of us having to update these docs every time we update the `api.json` schema

```suggestion
|                          // metadata about the installed SDKs.
```

docs/how-it-works.md
```diff
@@ -22,23 +22,28 @@ The `.api/` directory is where the CLI installation process stores all of its in
...
 └── apis/
+    ├── hoot/
     ├── readme/
-    |   ├── node_modules/
-    │   ├── index.js
-    │   ├── index.d.ts   // All types for the SDK, ready to use in your IDE.
+    │   ├── dist/          // The compiled source for your SDK. For JS languages
+    |   |                  // we offer CJS and ESM compatible builds.
```
kanadgupta, line 33: i think saying we have multiple builds might imply that they're separate SDKs? what do you think of this instead?

```suggestion
    │   ├── dist/          // The compiled source for your SDK. For JS,
    |   |                  // we offer a single SDK that's compatible with both ESM and CJS.
```

docs/how-it-works.md
```diff
@@ -47,13 +52,16 @@ The `api.json` file within `.api/` is where the CLI keeps track of everything th
 
 ```json
 {
-  "version": "1.0",
+  "$schema": "https://unpkg.com/api@7/schema.json",
   "apis": [
     {
+      "private": true,
       "identifier": "developers",
       "source": "@developers/v2.0#nysezql0wwo236",
```
kanadgupta, line 60: just b/c I'd rather the example be non-readme specific

```suggestion
      "identifier": "petstore",
      "source": "@petstore/v1.0#tl1e4kl1cl8eg8",
```

docs/how-it-works.md
kanadgupta, line 1: sorry for all the feedback in this file — i definitely plan on doing a full comb through our docs before we release v7

packages/api/src/commands/uninstall.ts
```diff
@@ -0,0 +1,94 @@
...
+  .addOption(new Option('-y, --yes', 'Automatically answer "yes" to any prompts printed'))
+  .action(async (identifier: string, options: Options) => {
+    // We don't know if we have `identifier` in the storage system yet, we just need to preload the
+    // system so we can access lockfiles.
+    const storage = new Storage('', SupportedLanguages.JS, identifier);
+
+    const entry = Storage.getFromLockfile(identifier);
+    if (!entry) {
+      logger(
+        `You do not appear to have ${identifier} installed. You can run \`npx api list\` to see what SDKs are present.`,
```
kanadgupta, line 30: this is great 🫶🏽

kanadgupta (commented):
mostly docs changes but this works great in my testing! if we're uninstalling the last remaining SDK, should we also remove the top-level `.api` and all of its contents or nah?

docs/how-it-works.md
```diff
@@ -22,23 +22,25 @@ The `.api/` directory is where the CLI installation process stores all of its in
...
+├── api.json               // The `package-lock.json` equivalent that records metadata about the
+|                          // installed SDKs.
 └── apis/
+    ├── hoot/
     ├── readme/
-    |   ├── node_modules/
-    │   ├── index.js
-    │   ├── index.d.ts   // All types for the SDK, ready to use in your IDE.
+    │   ├── dist/          // The compiled source for your SDK. For JS, we offer a single SDK that's
+    |   |                  // compatible with both ESM and CJS.
```
kanadgupta, line 31:
```suggestion
    |   |                  // fully typed and compatible with both ESM and CJS.
```

kanadgupta (approved):
one more tiny thing — otherwise LGTM!

PR #777, opened 2023-10-20
refactor(lockfile): consolidate TS and JSON schema type definitions

kanadgupta:
## 🧰 Changes

This does a bit of quality-of-life work to consolidate our TS types and our JSON Schema definition for the `api` lockfile schema. We now rebuild the `schema.json` file every time we run `npm version`, similar to how we do for `packageInfo.ts`.

## 🧬 QA & Testing

Does the `buildVersionedFiles.ts` work as expected when you run it locally? If the TS build and all the tests pass, we should be in good shape.


packages/api/schema.json
```diff
@@ -70,7 +62,8 @@
           "description": "The date that this SDK was last rebuilt or updated.",
           "examples": ["2023-10-19T20:35:39.268Z"]
         }
-      }
+      },
+      "additionalProperties": false
```
kanadgupta, line 66: this caught my attention thanks to `json-schema-to-ts` 👑

packages/api/schema.json
```diff
@@ -6,16 +6,8 @@
...
-    },
-    "apis": {
-      "type": "array",
-      "description": "The list of installed APIs",
-      "items": {
-        "$ref": "#/definitions/api"
-      }
-    }
+    "$schema": { "type": "string" },
+    "apis": { "type": "array", "description": "The list of installed APIs", "items": { "$ref": "#/definitions/api" } }
```
kanadgupta, line 10: prettier 🤷🏽‍♂️

packages/api/package.json
```diff
@@ -9,13 +9,14 @@
   "scripts": {
     "attw": "attw --pack --format table-flipped",
     "build": "tsc",
+    "build:versionedfiles": "NODE_OPTIONS=--no-warnings node --loader ts-node/esm bin/buildVersionedFiles.ts",
```
kanadgupta, line 12: i don't love this file name nor its corresponding `package.json` script name, open to ideas here 😬

packages/api/src/codegen/factory.ts
```diff
@@ -4,9 +4,14 @@ import type Oas from 'oas';
 
 import TSGenerator from './languages/typescript/index.js';
 
-export enum SupportedLanguages {
-  JS = 'js',
-}
+export const SupportedLanguages = {
```
kanadgupta, line 7: we can't use TS enums here anymore since we're deriving our supported languages from the JSON schema

packages/api/src/lockfileSchema.ts
```diff
@@ -0,0 +1,84 @@
...
+    installerVersion: {
+      type: 'string',
+      description: 'The version of `api` that was used to install this SDK.',
+      examples: ['7.0.0'],
+    },
+    integrity: {
+      type: 'string',
+      description:
+        'An integrity hash that will be used to determine on `npx api update` calls if the API has changed since the SDK was last generated.',
+      examples: ['sha512-ld+djZk8uRWmzXC+JYla1PTBScg0NjP/8x9vOOKRW+DuJ3NNMRjrpfbY7T77Jgnc87dZZsU49robbQfYe3ukug=='],
```
erunion, line 28: I'm not sure what the SHA I had in here was for but this one is the SHA for the Petstore spec.

```suggestion
      examples: ['sha512-otRF5TLMeDczSJlrmWLNDHLfmXg+C98oa/I/X2WWycwngh+a6WsbnjTbfwKGRU5DFbagOn2qX2SRvtBGOBRVGg=='],
```

packages/api/src/lockfileSchema.ts
```diff
@@ -0,0 +1,84 @@
...
+    },
+    language: {
+      type: 'string',
+      description: 'The language that this SDK was generated for.',
+      default: 'js',
+      enum: ['js'],
```
erunion, line 34: Now that this isn't JSON we can use the `SupportedLanguages` const we've got:

```suggestion
      default: SupportedLanguages.JS,
      enum: Object.values(SupportedLanguages),
```

Screen Shot 2023-10-20 at 10 08 15 AM

packages/api/src/lockfileSchema.ts
```diff
@@ -0,0 +1,84 @@
+import type { FromSchema } from '@readme/api-core/lib';
```
erunion, line 1:
```suggestion
import type { FromSchema } from '@readme/api-core/lib';

import { SupportedLanguages } from './codegen/factory.js';
```

packages/api/src/codegen/factory.ts
```diff
@@ -4,9 +4,14 @@ import type Oas from 'oas';
...
+  JS: 'js',
+} as const;
+
+type ObjectValues<T> = T[keyof T];
+
+export type SupportedLanguageType = ObjectValues<typeof SupportedLanguages>;
```
erunion, line 13: I don't think we have any TS types that end in `Type` anywhere else here so may as well be consistent.

```suggestion
export type SupportedLanguage = ObjectValues<typeof SupportedLanguages>;
```

erunion (changes requested):
some very small changes but otherwise this looks great.

packages/api/src/lockfileSchema.ts
kanadgupta: oooh smart!

packages/api/src/commands/install.ts
```diff
@@ -22,9 +24,9 @@ cmd
   )
   .addOption(new Option('-y, --yes', 'Automatically answer "yes" to any prompts printed'))
   .action(async (uri: string, options: { identifier?: string; lang: string; yes?: boolean }) => {
```
erunion, line 26:
```suggestion
  .action(async (uri: string, options: { identifier?: string; lang: SupportedLanguage; yes?: boolean }) => {
```

packages/api/src/commands/install.ts
```diff
@@ -22,9 +24,9 @@ cmd
...
-    let language: SupportedLanguages;
+    let language: SupportedLanguage;
     if (options.lang) {
-      language = options.lang as SupportedLanguages;
+      language = options.lang as SupportedLanguage;
```
erunion, line 29:
```suggestion
      language = options.lang;
```

erunion (approved):

PR #778, opened 2023-10-20
ci: cleanup

kanadgupta:
## 🧰 Changes

Node v18 is bundled with npm@9 so we don't need to install it ourselves. Also did a little bit of other cleanup in our `package.json` scripts etc.

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -11,10 +11,10 @@
     "build": "tsc",
     "debug:bin": "NODE_OPTIONS=--no-warnings node --loader ts-node/esm src/bin.ts",
     "lint:types": "tsc --noEmit",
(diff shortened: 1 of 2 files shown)


packages/api/package.json
```diff
@@ -11,10 +11,10 @@
     "build": "tsc",
     "debug:bin": "NODE_OPTIONS=--no-warnings node --loader ts-node/esm src/bin.ts",
     "lint:types": "tsc --noEmit",
-    "prebuild": "rm -rf dist/; npm run version",
+    "prebuild": "rm -rf dist/ && npm run version",
```
kanadgupta, line 14: semicolons are pretty problematic — if the preceding command fails then it moves onto the next one 😬

erunion (approved):

PR #781, opened 2023-10-20
fix(core): funky type issues with the main export exporting multiple things

erunion:
## 🧰 Changes

We've got a weird issue in the `api@next` dist where the `json-schema-to-ts` refactor I did in https://github.com/readmeio/api/pull/764 is mucking up some types. I'm not totally sure why!

`.cjs`, `.mts`, and `.ts` generated SDK:

cleanshot_2023-10-20_at_15 12 36_2x_360

However a `.mjs` and `.js` file have the types completely different:

cleanshot_2023-10-20_at_15 13 14_2x

While debugging why this was happening I noticed that `attw` was reporting that the types for `core` were all messed up:

before

After refactoring `ConfigOptions`, `FetchResponse` and `HTTPMethodRange` types into a new `@readme/api-core/types` file `attw` is now happy:

after

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -17,6 +17,10 @@
       "require": "./dist/lib/index.cjs",
       "import": "./dist/lib/index.js"
     },
+    "./types": {
+      "require": "./dist/types.d.cts",
+      "import": "./dist/types.d.ts"
+    },
     "./package.json": "./package.json"
   },
   "main": "dist/index.cjs",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -18,8 +18,8 @@
       "import": "./dist/lib/index.js"
     },
     "./types": {
-      "require": "./dist/types.d.cts",
-      "import": "./dist/types.d.ts"
+      "require": "./dist/types.d.cjs",
+      "import": "./dist/types.d.js"
     },
     "./package.json": "./package.json"
   },
--- a/packages/api/src/codegen/languages/typescript/index.ts
+++ b/packages/api/src/codegen/languages/typescript/index.ts
@@ -236,7 +236,7 @@ export default class TSGenerator extends CodeGenerator {
       sdkSource
         .getImportDeclarations()
         .find(id => id.getText().includes('HTTPMethodRange'))
-        ?.replaceWithText("import type { ConfigOptions, FetchResponse } from '@readme/api-core';");
+        ?.replaceWithText("import type { ConfigOptions, FetchResponse } from '@readme/api-core/types';");
     }
 
     return [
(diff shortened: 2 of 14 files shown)


erunion:
So this fix doesn't seem to totally fix the weird typing issues that we were seeing and seems to have broken it **everywhere** now into that weird `readonly` type state. I'm able to get it working everywhere with this diff:

```diff
diff --git a/packages/core/package.json b/packages/core/package.json
index 0f4c2d4..6f61fb7 100644
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -23,6 +23,13 @@
     },
     "./package.json": "./package.json"
   },
+  "typesVersions": {
+    "*": {
+      "errors/fetchError": ["./dist/errors/fetchError.d.ts"],
+      "lib": ["./dist/lib/index.d.ts"],
+      "types": ["./dist/types.d.ts"]
+    }
+  },
   "main": "dist/index.cjs",
   "module": "dist/index.js",
   "types": "dist/index.d.cts",
```

Screen Shot 2023-10-20 at 4 12 30 PM

Screen Shot 2023-10-20 at 4 12 37 PM

Screen Shot 2023-10-20 at 4 12 41 PM

And running every variation works:

Screen Shot 2023-10-20 at 4 14 26 PM

`attw` is happy now too:

Screen Shot 2023-10-20 at 4 15 01 PM

I know we were down on `typesVersions` in https://github.com/readmeio/oas/pull/820 but maybe that's what we should start doing?

packages/core/package.json
```diff
@@ -17,6 +17,10 @@
       "require": "./dist/lib/index.cjs",
       "import": "./dist/lib/index.js"
     },
+    "./types": {
+      "require": "./dist/types.d.cts",
+      "import": "./dist/types.d.ts"
```
kanadgupta, line 22: my thinking is we should point to the JS files instead since I would imagine that using `.*ts` files here will muck up Node's module resolution. i think `.*js` is the move since `tsup` is properly generating those files (even though they're empty) and [modern TS should pick up on the co-located declaration files](https://www.typescriptlang.org/docs/handbook/modules/reference.html#example-subpaths-conditions-and-extension-substitution).

```suggestion
      "require": "./dist/types.d.cjs",
      "import": "./dist/types.d.js"
```

kanadgupta (approved):
re: [this comment](https://github.com/readmeio/api/pull/781#issuecomment-1773510008), i'm still skeptical of the `typesVersions` approach — based on what i've read about it, our usage here is against the grain of what `typesVersions` is designed for (and this is especially the case in https://github.com/readmeio/oas/pull/820, where i'm confident we're using it incorrectly). i'm open to trying it out here though, mind opening a separate PR with those changes?

i think the changes in this branch look good (minus the one suggestion i noted below) — mind tagging a release with this PR? i'm having trouble getting it working properly in my sandbox environment.

PR #782, opened 2023-10-23
feat: tsconfig, gitignore, npm script enhancements

kanadgupta:
| 🚥 Resolves RM-8234, Resolves RM-8189 |
| :------------------- |

## 🧰 Changes

Makes a few enhancements in the codegen'd SDK directory:

- [x] Makes a few changes to the `tsconfig.json` to better match what's going on [(this cheatsheet has been very helpful)](https://www.totaltypescript.com/tsconfig-cheat-sheet)
- [x] Creates a `.gitignore` file that ignores the `dist/` directory since that directory is being autogenerated on a regular basis with the `prepare` npm script
- [x] Adds a `lint` npm script so the user can typecheck their SDK if needed.

## 🧬 QA & Testing

I confirmed that the SDK changes work well in [my sandbox](https://github.com/readmeio/tmp-api-tsup-sandbox)!


packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -419,6 +420,22 @@ sdk.server('https://eu.api.example.com/v14');`),
     return sourceFile;
   }
 
+  /**
+   * Creates a `.gitignore` file to prevent the `dist/` directory from being tracked.
+   *
+   */
+  createGitIgnore() {
+    const file = `# This file prevents the \`dist/\` directory from being tracked via git.
```
erunion, line 428: how do you feel about pulling this and the README file template out into isolated files that we do an `fs.readFileSync` on here? maybe put them in `src/codegen/languages/templates`?

packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -428,8 +445,11 @@ sdk.server('https://eu.api.example.com/v14');`),
 
     const config: TsConfigJson = {
       compilerOptions: {
-        module: 'NodeNext',
+        esModuleInterop: true,
+        module: 'ESNext',
+        moduleResolution: 'Bundler',
         resolveJsonModule: true,
+        strict: true,
```
erunion, line 452: kind of surprised that the sdks we're generating are passing strict mode

packages/test-utils/sdks/alby/tsconfig.json
```diff
@@ -1,7 +1,10 @@
 {
   "compilerOptions": {
-    "module": "NodeNext",
-    "resolveJsonModule": true
+    "esModuleInterop": true,
+    "module": "ESNext",
+    "moduleResolution": "Bundler",
+    "resolveJsonModule": true,
+    "strict": true
```
erunion, line 7: is `attw` still happy with these changes?

packages/api/src/codegen/languages/typescript/index.ts
kanadgupta: the gitignore feels TS-specific, right? and i was actually thinking of adding some docs specific to TS in the README as well

packages/test-utils/sdks/alby/tsconfig.json
kanadgupta: good reminder, i hadn't checked that — but yes! 

CleanShot 2023-10-25 at 13 55 04@2x

packages/api/src/codegen/languages/typescript/index.ts
erunion: ah yeah that's fair. want to at least pull them out into a TS-specific template directory? might be easier having them as flatfiles so you don't have to escape backticks.

packages/api/src/codegen/languages/typescript/index.ts
kanadgupta: :cry-party:

packages/api/src/codegen/languages/typescript/index.ts
kanadgupta: happy to do that with the gitignore file! but honestly since we use a [decent amount of templated data](https://github.com/readmeio/api/blob/08925bc4c262ed9d3a4492b7e6e385d5276a499c/packages/api/src/codegen/languages/typescript/index.ts#L665-L667) in the README file i don't think we gain much by moving it to a separate file and then setting up find/replace logic

erunion: k that's cool to leave it alone

erunion (approved):

PR #783, opened 2023-10-24
feat(api): improvements to the installation screen

erunion:
| 🚥 Resolves RM-8185 |
| :------------------- |

## 🧰 Changes

This work is some QOL improvements to the post-installation screen where we'll now, using a very simple algorithm to determine the best endpoint for a demo[^1], show a code example to folks so they can immediately get started with their newly generated SDK. Here's what it looks like:

Screen Shot 2023-10-23 at 5 02 24 PM

And in the case of us not being able to find a good endpoint to demo[^2] we just won't show a code example:

Screen Shot 2023-10-23 at 5 02 43 PM

[^1]: With the exception of us being able to handle auth examples, this algorithm is the same that powers the "recommended" API endpoint section of ReadMe Realtime configuration.
[^2]: In this case this API definition was only comprised endpoints with path parameters.

### DIFF
--- a/packages/api/src/commands/install.ts
+++ b/packages/api/src/commands/install.ts
@@ -1,5 +1,6 @@
 import type { SupportedLanguage } from '../codegen/factory.js';
 
+import chalk from 'chalk';
 import { Command, Option } from 'commander';
 import figures from 'figures';
 import Oas from 'oas';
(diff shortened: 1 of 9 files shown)


packages/api/src/commands/install.ts
```diff
@@ -23,19 +90,8 @@ cmd
...
-      ({ value: language } = await promptTerminal({
-        type: 'select',
-        name: 'value',
-        message: 'What language would you like to generate an SDK for?',
-        choices: [{ title: 'JavaScript', value: SupportedLanguages.JS }],
-        initial: 1,
-      }));
-    }
+  .action(async (uri: string, options: Options) => {
+    const language = await getLanguage(options);
```
erunion, line 94: Just some very small refactors to move this `language` and `identifier` selection out so this command action is a little easier to read.

packages/api/src/lib/suggestedOperations.ts
```diff
@@ -0,0 +1,118 @@
...
+  // `@readme/oas-to-snippet`, which would handle both, because we don't need the entire
+  // `oas-to-snippet` for what we're doing here. All we want to do is generate a very simple code
+  // example for `api` snippets and because we're controlling which kinds of endpoints we're
+  // generating this for the HAR dataset we're working with here is mostly a fully known object.
+  const har = oasToHar(oas, operation, undefined, auth)?.log?.entries?.[0]?.request;
+  if (!har) {
+    return false;
+  }
+
+  const snippet = client.convert(
```
erunion, line 93: I don't love this but I also **really** don't want to have to make `@readme/oas-to-snippet` a dependency because that depends on `httpsnippet-client-api` and if we need to make fixes to this snippet generation work it's be a multistep process:

1. publishing `httpsnippet-client-api`
2. pulling that release into `@readme/oas-to-snippet`
3. publishing `@readme/oas-to-snippet`.
4. puling that release into `api`
5. publishing `api`.

packages/httpsnippet-client-api/src/index.ts
```diff
@@ -47,10 +47,10 @@
       auth.push(`'${token.replace(/'/g, "\\'")}'`);
     });
 
-    return `sdk.auth(${auth.join(', ')});`;
+    return `${sdkVariable}.auth(${auth.join(', ')});`;
   }
 
-  return `sdk.auth('${authKey.replace(/'/g, "\\'")}');`;
+  return `${sdkVariable}.auth('${authKey.replace(/'/g, "\\'")}');`;
```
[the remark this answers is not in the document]
erunion: we own these `sdkVariable` and `api` identifier strings so this vulnerability won't happen.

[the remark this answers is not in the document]
erunion: we own these `sdkVariable` and `api` identifier strings so this vulnerability won't happen.

packages/httpsnippet-client-api/src/index.ts
```diff
@@ -47,10 +47,10 @@ function buildAuthSnippet(authKey: string | string[]) {
       auth.push(`'${token.replace(/'/g, "\\'")}'`);
     });
 
-    return `sdk.auth(${auth.join(', ')});`;
+    return `${sdkVariable}.auth(${auth.join(', ')});`;
```
erunion, line 50: found a bug in these `.auth()` snippets with custom variables while building this out.

kanadgupta (changes requested):
changes look good, really liking some of the refactors here! one small wrinkle: if the user's identifier has a hyphen, we should probably convert it to camelcase (or underscores, whatever you think makes the most sense!)

```diff
- import transit-new from '@api/transit-new';
+ import transitNew from '@api/transit-new';

- transit-new.getNextripAgencies()
+ transitNew.getNextripAgencies()
  .then(({ data }) => console.log(data))
  .catch(err => console.error(err));
```

kanadgupta (approved):

packages/httpsnippet-client-api/src/index.ts
```diff
@@ -143,10 +145,16 @@ const client: Client<APIOptions> = {
...
+    let sdkVariable: string;
     if (opts.identifier) {
       sdkPackageName = opts.identifier;
-      sdkVariable = opts.identifier;
+
+      sdkVariable = camelCase(opts.identifier);
+      if (isReservedOrBuiltinsLC(sdkVariable)) {
+        // If this identifier is a reserved JS word then we should prefix it with an underscore so
+        // this snippet can be valid code.
+        sdkVariable = `_${sdkVariable}`;
```
kanadgupta, line 156: ***great*** call here

erunion: yeah we do something similar for method and type names in codegen

PR #786, opened 2023-10-25
feat(api): creating LICENSE files and and SDK headings

erunion:
| 🚥 Resolves RM-8225 |
| :------------------- |

## 🧰 Changes

* [x] Creates a new `LICENSE` file in the codegen'd SDK directory if we're able to recognize a license in their OpenAPI definition `info.license` object.
* [x] Adding a `license` property to the `package.json` file if we have a recognized license.
* [x] Fill out a docblock on the main `SDK` class containing various pieces of data from their OpenAPI definition `info` object including: `title`, `description`, and `termsOfService`

This is what it all looks like on a fully specced out definition:

Screen Shot 2023-10-24 at 5 02 24 PM


packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -544,6 +547,84 @@ Generated at ${createdAt}
...
+   * Fill out a JSDoc heading on the SDK we're creating based on various pieces of data that may be
+   * present in the OpenAPI definition `info` object.
+   *
+   * Additionally if a license in `info.license` is recognized by the SPDX we'll also create a
+   * `LICENSE` file in for their generated SDK. We're only supporting SPDX-recognized licenses for
+   * this because we otherwise have no idea what the license represents.
+   *
+   * @see {@link https://spdx.org/licenses/}
+   */
+  private createSDKHeading() {
```
erunion, line 560: Might be nice to also support `info.contact` but there's not really a JSDoc equivalent for that.

packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -544,6 +547,84 @@ Generated at ${createdAt}
...
+
+        if (infoObject?.license?.url) {
+          docblock = TSGenerator.#addTagToDocblock(docblock, {
+            tagName: 'see',
+            text: `{@link ${infoObject.license.url}}`,
+          });
+        }
+
+        // We should add a LICENSE file to their SDK too.
+        this.project.createSourceFile('LICENSE', getLicense(spdxLicense));
```
erunion, line 619: Also wondering if we should set a `license` property in the `package.json` file at the same time.

packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -510,6 +514,10 @@ sdk.server('https://eu.api.example.com/v14');`),
       tsup: tsupOptions as JsonObject,
     };
 
+    if (!this.spdxLicense) {
+      delete pkg.license;
```
erunion, line 518: Doing this instead of adding `license` to `pkg` here because I like having `license` slotted into the object where it is instead of *after* `tsup`. 💅

packages/test-utils/sdks/petstore/src/index.ts
```diff
@@ -3,6 +3,18 @@ import type { ConfigOptions, FetchResponse } from '@readme/api-core/types';
...
+ * Swagger Petstore
+ *
+ * This is a sample server Petstore server.  You can find out more about Swagger at
+ * [http://swagger.io](http://swagger.io) or on [irc.freenode.net,
+ * #swagger](http://swagger.io/irc/).  For this sample, you can use the api key
+ * `special-key` to test the authorization filters.
+ *
+ * @see {@link http://swagger.io/terms/ Terms of Service}
+ * @license Apache-2.0
+ * @see {@link http://www.apache.org/licenses/LICENSE-2.0.html}
```
erunion, line 16: Figured it'd be nice to add this link to the license if they have `info.license.url` but I'm indifferent on it. Might also be nice that if their `info.license.name` and `info.license.identifier` aren't recognized but they have a URL that we could still at least add that license URL in here?

packages/test-utils/sdks/petstore/src/index.ts
```diff
@@ -3,6 +3,18 @@ import type { ConfigOptions, FetchResponse } from '@readme/api-core/types';
...
+ */
 class SDK {
```
erunion, line 18: One annoying thing about these docblocks is that they don't really transfer to our use-case:


https://github.com/readmeio/api/assets/33762/407ad41c-b9cd-47b5-8d3f-4c02c11c3c6f

The only way to see them is if you have an original instance of the SDK class, which we don't ever expose through codegen. I still like having the docblock there are pure documentation but from an implementation point of view it kind of sucks.

packages/test-utils/sdks/petstore/LICENSE
kanadgupta, line 1: very glad we have a fixture with a proper license defined!

kanadgupta (changes requested):

packages/test-utils/sdks/petstore/src/index.ts
kanadgupta: ah good point, maybe we can just flag that as a warning then? maybe our general spec validation process should check for valid SPDX identifiers too

packages/test-utils/sdks/petstore/src/index.ts
erunion: Should we add it in both spots? Or maybe. Move the SDK class into a new file and have the main index.ts file be the createSDK export?

packages/api/src/storage.ts
```diff
@@ -83,15 +83,24 @@ export default class Storage {
    * Reset the state of the entire storage system.
    *
    * This will completely destroy the contents of the `.api/` directory!
+   *
+   * @internal
    */
   static async reset() {
+    const noop = () => {
+      // If any of these file and directory removals fail we don't need to throw any errors because
+      // this method is only used internally in unit tests.
```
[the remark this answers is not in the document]
erunion: This method only exists for test teardowns because it completely destroys the .api directory. I'm comfortable with this silence here.

packages/test-utils/sdks/petstore/LICENSE
```diff
@@ -0,0 +1,97 @@
...
+
+APPENDIX: How to apply the Apache License to your work.
+
+To apply the Apache License to your work, attach the following boilerplate notice, with the fields enclosed by brackets "[]" replaced with your own identifying information. (Don't include the brackets!) The text should be enclosed in the appropriate comment syntax for the file format. We also recommend that a file or class name and description of purpose be included on the same "printed page" as the copyright notice for easier identification within third-party archives.
+
+Copyright [yyyy] <author>
```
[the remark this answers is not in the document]
erunion: I tried filling in the year as "yyyy" and "year" but got nowhere. Will take another look today.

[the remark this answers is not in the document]
kanadgupta: woof. turns out the `replacements` object only works with placeholders with the format `<key>` and not `[key]` 🤦🏽 

https://github.com/Ovyerus/license/blob/49df1c99e957764a6320a6a81f55c0dab1040310/src/getLicense.ts#L29

packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -501,6 +504,7 @@ sdk.server('https://eu.api.example.com/v14');`),
             }
           : {}),
       },
+      license: this.spdxLicense ?? '',
```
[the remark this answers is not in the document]
erunion: `license` can't be set to `undefined` in the `PackageJSON` type.

packages/api/package.json
```diff
@@ -52,6 +52,7 @@
     "figures": "^5.0.0",
     "httpsnippet-client-api": "file:../httpsnippet-client-api",
     "js-yaml": "^4.1.0",
+    "license": "^1.0.3",
```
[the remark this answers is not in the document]
erunion: agreed, there's unfortunately no other package on NPM like it and i don't really want to pull yet another package under my wings 🙃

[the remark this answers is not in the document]
kanadgupta: very fair lol

packages/test-utils/sdks/petstore/LICENSE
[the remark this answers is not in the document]
erunion: i'll file an issue and add a hack in our `getLicense()` call to manually replace that.

packages/test-utils/sdks/petstore/src/index.ts
kanadgupta: works for me! this might be a good idea for the publishing flow, i'll look into it. i'd argue that `@readme/openapi-parser` should eventually throw errors for invalid SPDX identifiers but not a priority by any means

erunion: yeah i'm just worried about folks using a license like "proprietary" which isn't recognized by the SPDX so we'd have to add in exclusions. feels like a losing battle but i also don't know what all different licenses we've got for specs in our DB

kanadgupta: yeah my rationale here is that [the spec](https://spec.openapis.org/oas/v3.1.0#license-object) says the `info.license.identifier` should be an SPDX expression, and if it's not SPDX then you should be using the `info.license.url` field. but yeah fair point re: the handling of existing specs — cool with whatever you decide here!

packages/api/test/codegen/languages/typescript/index.test.ts
```diff
@@ -97,22 +97,16 @@ function assertSDKFixture(file: string, fixture: string) {
 
 describe('typescript', () => {
   beforeEach(() => {
-    vi.setSystemTime(new Date('2023'));
```
erunion, line 100: Had to change this to a fuller date because `new Date('2023').getFullYear()` returns 2022.

packages/test-utils/sdks/operationid-quirks/src/index.ts
```diff
@@ -106,6 +106,10 @@ class SDK {
   }
 }
 
+/**
+ * Quirky `operationID` cases
+ *
+ */
 const createSDK = (() => { return new SDK(); })();
```
erunion, line 113: Moving these docblocks from the `SDK` class to this export caused them to get surfaced:

screen_shot_2023-10-25_at_10 02 38_am

packages/test-utils/sdks/petstore/LICENSE
```diff
@@ -0,0 +1,97 @@
...
+
+APPENDIX: How to apply the Apache License to your work.
+
+To apply the Apache License to your work, attach the following boilerplate notice, with the fields enclosed by brackets "[]" replaced with your own identifying information. (Don't include the brackets!) The text should be enclosed in the appropriate comment syntax for the file format. We also recommend that a file or class name and description of purpose be included on the same "printed page" as the copyright notice for easier identification within third-party archives.
+
+Copyright [2023]
```
erunion, line 79: The petstore spec doesn't have `info.contact` so we can't fill in an `author` in here.

packages/api/src/codegen/codegenerator.ts
```diff
@@ -49,6 +56,36 @@ export default abstract class CodeGenerator {
...
+      // we'll run the risk of licensing it under a license they didn't intend.
+      if (findLicense(spdxLicense).length === 1) {
+        this.spdxLicense = spdxLicense;
+      }
+    }
+
+    if (infoObject?.contact) {
+      if (infoObject.contact?.name && infoObject.contact?.email) {
+        this.apiContact.name = `${infoObject.contact.name} <${infoObject.contact.email}>`;
+      }
```
kanadgupta, line 83: thoughts on this?

```suggestion
      if (infoObject.contact?.name) {
        this.apiContact.name = `${infoObject.contact.name}`;
        if (infoObject.contact?.email) {
          this.apiContact.name += ` <${infoObject.contact.email}>`;
        }
      }
```

packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -392,9 +394,108 @@ sdk.server('https://eu.api.example.com/v14');`),
...
+
+      // Some licenses like `Apache-2.0` have the year as `[yyyy]` but the `license` package only
+      // supports templating on `<templatekey>` so we need to fix this ourselves.
+      // https://github.com/Ovyerus/license/issues/12
+      const year = new Date().getFullYear().toString();
+      const license = getLicense(this.spdxLicense, {
+        year,
+        // `license` doesn't support empty strings so we need to fake it here
+        author: this.apiContact.name ?? ' ',
+      }).replace('yyyy', year);
```
kanadgupta, line 476: wanna make this regex so `[yyyy]` and `<yyyy>` are properly handled?

kanadgupta (commented):
a few tiny nits but otherwise LGTM!

kanadgupta (approved):

PR #787, opened 2023-10-25
fix(core): refactoring  `FromSchema` out of core

erunion:
| 🚥 Resolves RM-8230 |
| :------------------- |

## 🧰 Changes

We've been having a lot of typing issues with `FromSchema`  since my refactor in https://github.com/readmeio/api/pull/764 so I'm reverting that work and making `json-schema-to-ts` an explicit dependency on codegen'd SDKs again.


kanadgupta (approved):
😔

PR #792, opened 2023-10-25
refactor(snippets): moving the library over to be an HTTPSnippet plugin

erunion:
## 🧰 Changes

With the work I did in https://github.com/readmeio/httpsnippet/releases/tag/8.1.0 this slightly refactors[^1] `httpsnippet-client-api` to now be a proper new HTTPSnippet plugin so it can now be installed as a one-liner:

```js
import { HTTPSnippet, addClientPlugin } from '@readme/httpsnippet';
import plugin from 'httpsnippet-client-api';

addClientPlugin(plugin);
```

Previously the way it was installed was as such:

```js
import { HTTPSnippet, addTargetClient } from '@readme/httpsnippet';
import plugin from 'httpsnippet-client-api';

addTargetClient('node', plugin);
```

This plugin refactor allows us to move the target that we're targeting, `node`, directly onto the client plugin. This refactor, and the addition of an `installation` message on the client will allow us to do some fun stuff in `@readme/oas-to-snippet` and accept plugins from outside sources.

[^1]: Very much a breaking change refactor but because we're still in beta land with this work I'm not going to bother versioning this to another major.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3512,9 +3512,9 @@
       "integrity": "sha512-/dBh9qw3QhJYqlGwt2I+KUP/lQ6nytdCx3aq+GpMUhibLHF3O7fwoowNcTwlbnwtyJ+TJYTIIrp3oVUlRNx3fA=="
     },
     "node_modules/@readme/httpsnippet": {
-      "version": "8.0.1",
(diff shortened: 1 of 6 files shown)


kanadgupta (approved):
i'm having a little trouble wrapping my head around how all these changes fit together lol but on the whole excited for this!

PR #793, opened 2023-10-25
fix(api): run `npm install` with `--ignore-scripts`

erunion:
## 🧰 Changes

The `prepare` script that was added to codegen'd `package.json` files in https://github.com/readmeio/api/pull/761 is being executed when we install that SDK resulting us in running the compilation step twice. Because I'd like to keep the dependency installation and SDK dist compilation processes separate we should run `npm install` with the `--ignore-scripts` option.

fixes RM-8243

### DIFF
--- a/packages/api/src/codegen/languages/typescript/index.ts
+++ b/packages/api/src/codegen/languages/typescript/index.ts
@@ -1,4 +1,5 @@
 import type { InstallerOptions } from '../../factory.js';
+import type { ExecaReturnValue } from 'execa';
 import type Oas from 'oas';
 import type Operation from 'oas/operation';
 import type { HttpMethods, SchemaObject } from 'oas/rmoas.types';
@@ -60,6 +61,22 @@ interface OperationTypeHousing {
  */
 const REF_PLACEHOLDER = '::convert::';
 const REF_PLACEHOLDER_REGEX = /"::convert::([a-zA-Z_$\\d]*)"/g;
+
+function handleExecSuccess(res: ExecaReturnValue<string>, opts: InstallerOptions = {}) {
+  if (opts.dryRun) {
+    (opts.logger ? opts.logger : logger)(res.command);
+    (opts.logger ? opts.logger : logger)(res.stdout);
+  }
+}
+
+function handleExecFailure(err: Error, opts: InstallerOptions = {}) {
+  if (opts.dryRun) {
+    (opts.logger ? opts.logger : logger)(err.message);
+    return;
+  }
+
+  throw err;
+}
 
 export default class TSGenerator extends CodeGenerator {
   project: Project;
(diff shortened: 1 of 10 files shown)


kanadgupta (changes requested):
so i added a little prompt after [this line](https://github.com/readmeio/api/blob/67d8e50825772744832f020446094940e5ca4774/packages/api/src/commands/install.ts#L202) to confirm that the `dist/` directory wasn't being built before the compilation step, but it looks like it's being built:

CleanShot 2023-10-26 at 09 34 26@2x

i did a little digging and was able to fix this by making this change [on this line](https://github.com/readmeio/api/blob/67d8e50825772744832f020446094940e5ca4774/packages/api/src/codegen/languages/typescript/index.ts#L148):

```diff
-    return execa('npm', [...npmInstall, installDir].filter(Boolean))
+    return execa('npm', [...npmInstall].filter(Boolean), { cwd: installDir })
```

erunion:
Is the api in that case added to the root package.json file? I think thats just going to run npm i install the api and not add it as a dependency

kanadgupta:
> I think thats just going to run npm i install the api and not add it as a dependency

ah shit you're right 😬 my hunch is that the `--ignore-scripts` flag only applies to scripts in the `package.json` where you're running `npm i` from and it runs all the lifecycle scripts in the subdirectory's `package.json`

kanadgupta:
another idea:

1. build the initial `package.json` _without_ a `prepare` script
2. add the `prepare` script during the compilation step using [`npm pkg set`](https://docs.npmjs.com/cli/v10/commands/npm-pkg)?

packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -197,23 +197,15 @@ export default class TSGenerator extends CodeGenerator {
...
-      .catch(err => {
-        if (opts.dryRun) {
-          (opts.logger ? opts.logger : logger)(err.message);
-          return;
-        }
-
-        throw err;
-      });
+      .then(res => handleExecSuccess(res, opts))
+      .catch(err => handleExecFailure(err, opts));
```
erunion, line 208: might be nice to write a small wrapper for `execa()` that handles these success and failure states instead of having to copy-paste this code everywhere but i don't really have the energy at the moment to try to figure out how to clone all of execa's parameter typing into a wrapper

packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -197,23 +197,15 @@ export default class TSGenerator extends CodeGenerator {
   async compile(storage: Storage, opts: InstallerOptions = {}): Promise<void> {
     const installDir = storage.getIdentifierStorageDir();
 
-    await execa('npx', ['tsup'], {
+    await execa('npm', ['pkg', 'set', 'scripts.prepare=tsup'])
```
kanadgupta, line 200:
```suggestion
    await execa('npm', ['pkg', 'set', 'scripts.prepare=tsup'], { cwd: installDir })
```

packages/api/src/codegen/languages/typescript/index.ts
```diff
@@ -61,6 +62,22 @@ interface OperationTypeHousing {
 const REF_PLACEHOLDER = '::convert::';
 const REF_PLACEHOLDER_REGEX = /"::convert::([a-zA-Z_$\\d]*)"/g;
 
+function handleExecSuccess(res: ExecaReturnValue<string>, opts: InstallerOptions = {}) {
```
kanadgupta, line 65: great refactor here 🫶🏽

kanadgupta (changes requested):
i was able to confirm it works as expected with the fix below!

packages/api/src/codegen/languages/typescript/index.ts
erunion: ah i knew i forgot something

kanadgupta (approved):

PR #794, opened 2023-10-26
feat: throw error for legacy dynamic `require` users

kanadgupta:
| 🚥 Resolves RM-8241 |
| :------------------- |

## 🧰 Changes

If users were to try and use the `const sdk = require("api")("url");` pattern that has since been removed from `api` v7, they'll see a confusing error like this:

```
node:internal/modules/cjs/loader:1080
  throw err;
  ^

Error: Cannot find module 'api'
Require stack:
- /Users/kanadg/Code/readmeio/tmp-api-tsup-sandbox/src/legacy-require.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1077:15)
    at Module._load (node:internal/modules/cjs/loader:922:27)
    at Module.require (node:internal/modules/cjs/loader:1143:19)
    at require (node:internal/modules/cjs/helpers:119:18)
    at Object.<anonymous> (/Users/kanadg/Code/readmeio/tmp-api-tsup-sandbox/src/legacy-require.js:1:17)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:86:12) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Users/kanadg/Code/readmeio/tmp-api-tsup-sandbox/src/legacy-require.js'
  ]
}

Node.js v18.18.2
```

This improves the error so it now looks like this:

```
/Users/kanadg/Code/readmeio/api/packages/api/legacy-require-handler.cjs:19
  throw new Error(`Oops! You're attempting to use a legacy usage pattern that no longer exists.
  ^

Error: Oops! You're attempting to use a legacy usage pattern that no longer exists.

You can read more about this in our docs: TKTK

`api` can now generate strongly-typed SDKs that work in pretty much every JS runtime!

Try running this command to get started:

npx api@latest install @metrotransit/v1.0#2w2je3tolbchzuph

    at legacyHandler (/Users/kanadg/Code/readmeio/api/packages/api/legacy-require-handler.cjs:19:9)
    at Object.<anonymous> (/Users/kanadg/Code/readmeio/tmp-api-tsup-sandbox/src/legacy-require.js:1:31)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:86:12)
    at node:internal/main/run_main_module:23:47

Node.js v18.18.2
```

I also added a little type declaration file so TS users will see this:

CleanShot 2023-10-26 at 10 51 51@2x

## 🧬 QA & Testing

### DIFF
--- a/packages/api/legacy-require-handler.cjs
+++ b/packages/api/legacy-require-handler.cjs
@@ -1,3 +1,4 @@
+// @ts-check
 /* eslint-disable import/no-commonjs */
 
 /**
--- a/packages/api/legacy-require-handler.cjs
+++ b/packages/api/legacy-require-handler.cjs
@@ -1,13 +1,6 @@
 /* eslint-disable import/no-commonjs */
 
 /**
- * You can regenerate the type declarations for this file by running the following:
- * ```sh
- * npx tsc legacy-require-handler.cjs --checkJs --declaration --emitDeclarationOnly
- * ```
- */
-
-/**
  * In `api` v6, this package was a JS library used JavaScript Proxies to dynamically generate SDKs at runtime.
  * This functionality has been sunset in `api` v7 to focus on its CLI, which builds strongly-typed SDKs.
  * This file is here to throw errors for users that attempt to use the legacy library.
--- a/packages/api/legacy-require-handler.cjs
+++ b/packages/api/legacy-require-handler.cjs
@@ -24,5 +24,6 @@ You can read more about this in our docs: TKTK
 
 Try running this command to get started:
 
-npx api@latest install ${input}`);
+npx api@latest install ${input}
+`);
 };
--- a/packages/api/legacy-require-handler.cjs
+++ b/packages/api/legacy-require-handler.cjs
@@ -1,0 +1,28 @@
+/* eslint-disable import/no-commonjs */
+
+/**
+ * You can regenerate the type declarations for this file by running the following:
+ * ```sh
+ * npx tsc legacy-require-handler.cjs --checkJs --declaration --emitDeclarationOnly
+ * ```
+ */
+
+/**
+ * In `api` v6, this package was a JS library used JavaScript Proxies to dynamically generate SDKs at runtime.
+ * This functionality has been sunset in `api` v7 to focus on its CLI, which builds strongly-typed SDKs.
+ * This file is here to throw errors for users that attempt to use the legacy library.
+ *
+ *
+ * @param {unknown} input
+ */
+module.exports = function legacyHandler(input) {
+  throw new Error(`Oops! You're attempting to use a legacy usage pattern that no longer exists.
+
+You can read more about this in our docs: TKTK
+
+\`api\` can now generate strongly-typed SDKs that work in pretty much every JS runtime!
+
+Try running this command to get started:
+
+npx api@latest install ${input}`);
+};
--- a/packages/api/legacy-require-handler.cjs
+++ b/packages/api/legacy-require-handler.cjs
@@ -16,6 +16,12 @@
  * @param {unknown} input
  */
 module.exports = function legacyHandler(input) {
+  let identifier = 'https://raw.githubusercontent.com/OAI/OpenAPI-Specification/main/examples/v3.0/petstore.json';
+
+  if (typeof input === 'string') {
+    identifier = input;
+  }
+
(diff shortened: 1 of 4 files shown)


packages/api/legacy-require-handler.cjs
```diff
@@ -0,0 +1,29 @@
...
+ * This functionality has been sunset in `api` v7 to focus on its CLI, which builds strongly-typed SDKs.
+ * This file is here to throw errors for users that attempt to use the legacy library.
+ *
+ *
+ * @param {unknown} input
+ */
+module.exports = function legacyHandler(input) {
+  throw new Error(`Oops! You're attempting to use a legacy usage pattern that no longer exists.
+
+You can read more about this in our docs: TKTK
```
kanadgupta, line 21: added a ticket to work on the migration guide and ticket it here: RM-8248

packages/api/legacy-require-handler.cjs
```diff
@@ -0,0 +1,29 @@
...
+ * ```
+ */
+
+/**
+ * In `api` v6, this package was a JS library used JavaScript Proxies to dynamically generate SDKs at runtime.
+ * This functionality has been sunset in `api` v7 to focus on its CLI, which builds strongly-typed SDKs.
+ * This file is here to throw errors for users that attempt to use the legacy library.
+ *
+ *
+ * @param {unknown} input
```
erunion, line 16:
```suggestion
 * @param {string} input
```

packages/api/legacy-require-handler.cjs
```diff
@@ -0,0 +1,29 @@
+/* eslint-disable import/no-commonjs */
+
+/**
+ * You can regenerate the type declarations for this file by running the following:
+ * ```sh
+ * npx tsc legacy-require-handler.cjs --checkJs --declaration --emitDeclarationOnly
```
erunion, line 6: want to add this as `build:legacy-handler` script that's run post `npm run build`?

erunion (approved):

packages/api/legacy-require-handler.cjs
kanadgupta: i thought about this but we technically allowed OAS documents back in the day:

https://github.com/readmeio/api/blob/709c54bdb8fd64672f3f06f165410afe7798713d/packages/api/src/index.ts#L201

and I kinda like how there's currently just one TS error rather than another one that's basically a red herring:

CleanShot 2023-10-27 at 16 00 54@2x

erunion: ah true. tbh i think that was only ever exposed for unit tests so i doubt anyone in the wild used it but leaving this as `unknown` is cool.

packages/api/legacy-require-handler.cjs
```diff
@@ -0,0 +1,29 @@
...
+module.exports = function legacyHandler(input) {
+  throw new Error(`Oops! You're attempting to use a legacy usage pattern that no longer exists.
+
+You can read more about this in our docs: TKTK
+
+\`api\` can now generate strongly-typed SDKs that work in pretty much every JS runtime!
+
+Try running this command to get started:
+
+npx api@latest install ${input}
```
erunion, line 27: if this is an object might not want to print it out.

packages/api/package.json
```diff
@@ -34,6 +37,7 @@
   },
   "files": [
     "dist",
+    "legacy-require-handler.**",
```
erunion, line 40: does the `files` array support globs like this?

erunion (approved):

packages/api/package.json
kanadgupta: CleanShot 2023-10-30 at 12 19 59@2x

yep, confirmed that `npm pack` includes it

PR #809, opened 2023-11-02
feat: add example code snippet to README

kanadgupta:
## 🧰 Changes

This takes our fantastic example code snippet function and reuses it so we can populate the codegen'd SDK's README.md with example usage. If an example doesn't exist, it falls back to the existing copy:

> Add SDK setup information and usage examples here so your users get started in a jiffy! 🚀

## 🧬 QA & Testing

Right now we call `getExampleCodeSnippet()` twice every time someone generates an SDK and I'm wondering if there's a better way for us to consolidate.

But otherwise, does everything look good?


packages/api/src/lib/suggestedOperations.ts
```diff
@@ -124,3 +124,12 @@ export async function buildCodeSnippetForOperation(oas: Oas, operation: Operatio
 
   return snippet;
 }
+
+export async function getExampleCodeSnippet(oas: Oas, identifier: string) {
```
kanadgupta, line 128: i moved this to here since it was an easy refactor, but i feel like it should live in the `CodeGenerator` class or somewhere where we're not running this code twice during codegen... any ideas @erunion?

erunion: Yeah this should probably be an instance method somewhere on the `CodeGenerator` class. Maybe it saves the result to a `codesnippet` property that we can re-retrieve if a snippet hasnt been made yet so we don't need to make it twice.

erunion (approved):
fantastic addition

PR #815, opened 2023-11-08
feat(api): import schemas in the types file as types

erunion:
## 🧰 Changes

Because all of the schemas that we're importing in codegen'd `types.ts` files are for types only we can import them explicitly as types.


kanadgupta (approved):
this appears to be a fine change according to my testing but 
i don't think i understand why... technically speaking, aren't the schemas in `schemas.ts` actually JSON schema objects as `const`s?

erunion:
they are but they're used in `types.ts` as only types

PR #816, opened 2023-11-08
feat(snippets): splitting apart package and variable name configs

erunion:
## 🧰 Changes

This splits apart some confusing work I did in `httpsnippet-client-api` where I was always tying the `@api/<package name>` to the variable name if a custom variable name was supplied. For example if you supplied an `identifier` of `petstore` and `variableName` of `buster`, the variable name would be `buster` but `@api/<package name>` would also be `@api/buster`. Very confusing and not really now we intend this custom variable name support to be handled.

Instead I've renamed the `identifier` option to the clearer `packageName` and added a new dedicated `variableName` config for custom variable names. This small refactor will require a couple small changes to `@readme/oas-to-snippet`.

### DIFF
--- a/packages/httpsnippet-client-api/src/index.ts
+++ b/packages/httpsnippet-client-api/src/index.ts
@@ -4,7 +4,6 @@ import type { Operation } from 'oas/operation';
 import type { HttpMethods, OASDocument } from 'oas/types';
 
 import { CodeBuilder } from '@readme/httpsnippet/helpers/code-builder';
-import camelCase from 'camelcase'; // eslint-disable-line import/no-extraneous-dependencies
 import contentType from 'content-type';
 import Oas from 'oas';
 import { matchesMimeType } from 'oas/utils';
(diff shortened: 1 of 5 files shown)


packages/httpsnippet-client-api/src/index.ts
```diff
@@ -151,20 +193,22 @@ const client: Client<APIOptions> = {
...
+    let sdkPackageName: string;
+    if (opts.api.packageName) {
+      sdkPackageName = opts.api.packageName;
+    } else {
+      const registryUUID = getProjectPrefixFromRegistryUUID(opts.api.registryURI);
+      sdkPackageName = registryUUID || 'company-name';
+    }
 
-      sdkVariable = camelCase(opts.api.identifier);
+    let sdkVariable: string = 'sdk';
```
kanadgupta, line 204: i feel like the SDK variable should fall back to the `sdkPackageName` if `variableName` isn't present as opposed to `sdk`, thoughts?

```suggestion
    let sdkVariable = sdkPackageName;
```

kanadgupta (approved):

packages/httpsnippet-client-api/src/index.ts
erunion: I thought that we wanted this custom variable name stuff to be completely custom for folks who want it but I'm ok with this.

packages/httpsnippet-client-api/src/index.ts
```diff
@@ -36,6 +35,42 @@ function stringify(obj: any, opts = {}) {
...
+ * Convert a string that might contain spaces or special characters to one that can safely be used
+ * as a TypeScript interface or enum name.
+ *
+ * This function has been adapted and slighty modified from `json-schema-to-typescript`. This
+ * function also exists in `api` TS codegen.
+ *
+ * @license MIT
+ * @see {@link https://github.com/bcherny/json-schema-to-typescript}
+ */
+function toSafeString(str: string) {
```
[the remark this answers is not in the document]
erunion: yep

packages/httpsnippet-client-api/src/index.ts
kanadgupta: agreed, i like the idea of a custom variable name! this is just in the fallback case

PR #819, opened 2023-11-08
test: run `tsc` on SDK fixtures

kanadgupta:
## 🧰 Changes

I've done cursory checks in the past to make sure our generated SDKs meet the requirements of TS strict mode, but it just occurred to me that we're not doing any type-checking in our tests, so this PR enables that. Glad to report that the SDKs passed with flying colors ✨

## 🧬 QA & Testing

Do tests still pass?

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -6,7 +6,6 @@
     "declaration": true,
     "declarationMap": true,
     "sourceMap": true,
-    "strict": true,
     "useUnknownInCatchVariables": false
   }
 }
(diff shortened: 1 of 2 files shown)


tsconfig.base.json
```diff
@@ -6,7 +6,6 @@
     "declaration": true,
     "declarationMap": true,
     "sourceMap": true,
-    "strict": true,
```
kanadgupta, line 9: this is already present in `@tsconfig/node18/tsconfig.json` so  I removed it!

erunion (approved):

Issue #604, opened 2023-01-12
Issue with NestJS


bennettl:
Trying to use this for my NestJS project, and when running got the following compilation error

import type * as types from './types';
^^^^^^

SyntaxError: Cannot use import statement outside a module

It seems like this sits outside the `src` folder, not sure if it can be an issue. Any help is appreciated!!!

erunion:
I'm not at all familiar with NestJS but I have some questions about your install.

1. How did you install your API SDK?
2. Did you install it as JS or TypeScript?
3. CJS or ESM?
4. How are you loading your API SDK?

murtinha:
Im also having this problem after adding **resolveJsonModule** to my ts.config to fix the json import from the generated .api folder.

Answering @erunion 
1. `npx api install @open-api-docs/MYAPI`
2. Typescript
3. CJS
4. importing it directly from the generated .api folder

erunion:
I don't use NestJS so I'm not sure how to solve this. Do you know why it's trying to use the TS file directly instead of compiling it into your app when you build your main output `main.js` file?

linzera:
Probably because the `.app` directory is not being included in the build phase.

murtinha:
Im not 100% sure but it seems to have to do with nest dependency injection

Cmoen11:
This is the same issue in next.js

erunion:
For what it's worth, if you're generating a JS SDK `api` will still generate TS types for you to use in local development. You don't need to generate a TS SDK to get TS types.

Cmoen11:
Alright, @erunion I think I've found out what's causing this. And it is neither next.js nor nuxt.js, it is for the project which uses yarn instead of npm. 

It seems like the lib is forcing on npm, and it's creating conflicting package-lock.json with the yarn.lock file. switching over to npm instead seems to solve my problem.

Cmoen11:
I think i've located the problem.

https://github.com/readmeio/api/blob/ebb90af3605cb522f4aafaba75b9c9fe581b6534/packages/api/src/cli/codegen/languages/typescript.ts#LL156C21-L156C21

So if you're using a different package manager you may stumble on some problems caused by this.

erunion:
Besides @Cmoen11, is everyone else in this thread using Yarn?

afreemanio:
Alright, so, I ran into this one today.
Can confirm, not using Yarn, not using Nest or Next or anything, just plain node.

[...]

PR #820, opened 2023-11-16
fix: use dynamic import for node-specific libraries

kanadgupta:
## 🧰 Changes

I went down a little rabbit hole to see how it would work to load these SDKs in a browser environment as part of the docs I want to write up. The closest I got to success was this error (even with a webpack config and various tweaks to it):

CleanShot 2023-11-16 at 10 31 50@2x

Looks like the same error alluded to here: https://github.com/readmeio/api/issues/604#issuecomment-1516682177

I figured that by only loading the node-specific libraries within the function that needs it, we might be able to successfully run this in a browser environment?

## 🧬 QA & Testing

I was unable to successfully `npm link` this into my setup in a way that played nicely with webpack so hard to see if this will work. Open to other ideas here!


kanadgupta:
Update: chatted with @erunion about this a while back and we might be able to bring in the `datauri` methods that we use directly — will pick this up when I have capacity again!

joaocanaverde-blue:
@kanadgupta @erunion Hi! Any chance this can be merged? Happy to contribute if anything is needed.

kanadgupta:
@erunion ready for review again. I tried playing around with vendoring `datauri` but it felt like overkill since we'd need to create a separate subpackage for it since it's used in both `core` and `testing-utils`

kanadgupta:
Hi there @joaocanaverde-blue — [`v7.0.0-beta.11`](https://github.com/readmeio/api/releases/tag/v7.0.0-beta.11) includes this change.

PR #854, opened 2024-02-08
feat: avoid modifying the response object when calling response.text()

franciscopantojag:
| 🚥 Resolves ISSUE_ID |
| :------------------- |

## 🧰 Changes

- Avoid mutating the response object when calling `response.text()`. This can be helpful if we want to consume the response in a different way, like `response.arrayBuffer()` when working with files

## 🧬 QA & Testing

- With this change, the original response object is returned.


kanadgupta (changes requested):
hi there! thanks for this change, it looks great. could you add a test? this is the file you'll want to add your test(s) to: https://github.com/readmeio/api/blob/main/packages/core/test/lib/parseResponse.test.ts

let us know if you need a hand, thanks!

franciscopantojag:
Hello @kanadgupta , test added!
Thanks for the suggestion 😊

kanadgupta (approved):
thanks so much again @franciscopantojag!

PR #855, opened 2024-02-12
fix: ensure prepareServer respects server variables when custom url is the same as the OAS url

yoitsro:
| 🚥 Resolves ISSUE_ID |
| :------------------- |

## 🧰 Changes

A truthy check was being performed on the `serverIdx`, which would effectively ignore the value when `serverIdx` was 0.

This fixes that to ensure that server variables are respected when the server URL passed in is the same as the server URL defined in the OAS.

## 🧬 QA & Testing

It now works for us, whereas it didn't before.


kanadgupta (approved):
thanks!

Issue #821, opened 2023-11-17
define the Type of the sdk object.


mewc:
Im trying to get an inferred type of the whole imported "sdk" object

for example.
```
import workspaces from '@api/generated-sdk-of-some-kind'

const configureApiModule = (apiModule: any ) => {
  apiModule.server(url, { basePath: 'v1' })
  apiModule.auth(username, password)
}

modules.forEach(configureApiModule)
```

i want to be able to make the apiModule typed properly, but i can't seem to export or use any of the `type SDKInstance = ReturnType<typeof createSDK>;` or InstanceOf since the `createSDK` call is an IIFE.

I also dont want to edit codegen stuff ofc. Am i missing anything?

erunion:
@kanadgupta Any ideas on how we can handle this? We can't export SDK from the generated `index.ts` files because that'd collide with the default export in CJS environments, and we can't do the same with an `export type APISDK = typeof SDK` in the same file for the same reason.

kanadgupta:
@erunion what if we defined the SDK class in a `sdk.ts` file and created an export in the `package.json` for that? And then the IIFE would remain in `index.ts` and import from `sdk.ts`?

erunion:
@kanadgupta That'd work

PR #856, opened 2024-02-13
feat(api): export SDK

yoitsro:
| 🚥 Resolves https://github.com/readmeio/api/issues/821 |
| :------------------- |

## 🧰 Changes

In some use cases, the SDK class needs to be instantiated several times.

For instance, in a multi-tenancy scenario, there may be multiple instances of the same API server, but serving different tenants' data from different base URLs. In this case, we need to instantiate multiple clients with different server URLs.

## 🧬 QA & Testing

The unit test fixtures have been updated accordingly to reflect that the SDK is now exported.


kanadgupta (changes requested):
hi, thanks so much for this contribution! unfortunately when compiling TS into CJS and ESM, files with both default and non-default exports typically lead to typing issues in certain environments. we don't have great test coverage on this front quite yet unfortunately.

what if the class was the default export from a new `src/sdk.ts` file and the IIFE in `src/index.ts` imports that?[^1]

[^1]: first discussed [here](https://github.com/readmeio/api/issues/821#issuecomment-1839242629)

yoitsro:
Yeah, I saw that discussion after submitting the PR but didn't have time to sort it. Let me take a quick look at it.

yoitsro:
Here are the changes. Once they're good, I'll squash the commits down (or I guess you could do that too if required).

kanadgupta (approved):
thanks again, this looks great!

kanadgupta:
> I'll squash the commits down (or I guess you could do that too if required).

and we'll take care of squashing when we merge, thanks!

erunion (approved):
thanks for this!

Issue #849, opened 2024-01-30
NPM Audit Report Vulnerability due to lodash.setwith


andreisaikouski:
As per title, `npm audit` report provides:

```
lodash.setwith  *
Severity: high
Prototype Pollution in lodash - https://github.com/advisories/GHSA-p6mc-m468-83gw
fix available via `npm audit fix --force`
Will install api@2.6.0, which is a breaking change
node_modules/lodash.setwith
  api  >=5.0.0-beta.0
  Depends on vulnerable versions of lodash.setwith
  node_modules/api
```

brickfungus:
I was able to get around it with this https://github.com/lodash/lodash/issues/5809#issuecomment-1910560681
but a fix here would be nice

andreisaikouski:
@brickfungus thanks for the recommend. Yeah I did the same, but wanted to raise an issue here as I saw one doesn't exist

jdleesmiller:
Thank you for merging the PR. Will there be a new release soon that includes the change?

erunion:
Hi, sorry for the delay on this one but I've just published `v7.0.0-beta.6` to our `next` channel.

lpgera:
Hey! I totally understand that releasing a stable v7 may take a lot of time, so is there any chance for backporting this security patch and releasing it as v6.1.2? I'd be happy to create a PR if there's support for this and someone creates a v6 branch at e440a28841e753895bfb5315b4d31f679824fa39 to serve as a base for it.

erunion:
Apologies for dragging ass on this, I've just backported the `lodash.<package>` -> `lodash` package migrations that happened https://github.com/readmeio/api/pull/859 down to our v6 series with a new v6.1.2 release on NPM. 

https://github.com/readmeio/api/compare/v6.1.1...v6.1.2

lpgera:
Much appreciated!

PR #859, opened 2024-02-23
chore(deps): use lodash instead of per method packages

jdleesmiller:
| 🚥 Resolves https://github.com/readmeio/api/issues/849 |
| :------------------- |

## 🧰 Changes

Use of `lodash` per method packages [is now discouraged (docs)](https://lodash.com/per-method-packages), and they appear to be unmaintained and have security vulnerabilities. As explained in the linked docs, depending on lodash is preferred.

## 🧬 QA & Testing

The tests on the `api` package passed locally.


jdleesmiller:
Thanks for approving the CI runs. It looks like a generic CI failure... which I am not sure what to do with. It is possible that I have broken something in lerna by not using lerna to remove the package?

kanadgupta:
Hi! See https://github.com/lodash/lodash/issues/5107, specifically this part:

> This doesn’t work with modules (.mjs files) in NodeJS. Instead, lodash needs to be imported as a default import.

The `api` subpackage (e.g., `./packages/api` is ESM-only. I tried swapping this out for `lodash-es` and it appears to work as expected.

kanadgupta (approved):
Thanks!

PR #898, opened 2024-06-25
feat: use project name rather than `sdk` in sdk snippet

darrenyong:
| 🚥 Resolves RM-6573 |
| :------------------- |

## 🧰 Changes
We wanted to update the `sdk` in our SDK snippet generation to use the site name so it feels more personalized. 

## 🧬 QA & Testing

### DIFF
--- a/packages/httpsnippet-client-api/src/index.ts
+++ b/packages/httpsnippet-client-api/src/index.ts
@@ -164,7 +164,7 @@ const client: Client<APIOptions> = {
       }
     } else {
       sdkPackageName = getProjectPrefixFromRegistryUUID(opts.api.registryURI);
-      sdkVariable = camelCase(sdkPackageName as string);
+      sdkVariable = camelCase(sdkPackageName || 'sdk');
     }
 
     const operationSlugs = foundOperation.url.slugs;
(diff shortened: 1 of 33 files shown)


packages/httpsnippet-client-api/src/index.ts
```diff
@@ -164,7 +164,7 @@ const client: Client<APIOptions> = {
       }
     } else {
       sdkPackageName = getProjectPrefixFromRegistryUUID(opts.api.registryURI);
-      sdkVariable = 'sdk';
+      sdkVariable = camelCase(sdkPackageName as string);
```
erunion, line 167: Probably don't want to suppress the TS error here that `sdkPackageName` might be `null` because that'll happen if `registryURI` isn't properly formulated.

```suggestion
      sdkVariable = camelCase(sdkPackageName || 'sdk');
```

erunion (approved):

packages/httpsnippet-client-api/src/index.ts
darrenyong: ah, good call!

gkoberger:
Yessss!

PR #910, opened 2024-07-03
fix: backport lodash upgrades to v6.1.1

erunion:
| 🚥 Resolves https://github.com/readmeio/api/issues/849 |
| :------------------- |

## 🧰 Changes

This backports the Lodash upgrades that happened in https://github.com/readmeio/api/pull/859 down to our v6 release series.


erunion:
Confirmed these tests are passing locally with the exception of the ones that are asserting a version string in generated SDKs. Safe to publish.

erunion:
v6.1.2 published

PR #1004, opened 2025-03-05
ci: running `attw` along our test suite

erunion:
## 🧰 Changes

This bundles `attw` into our build processes for ensuring that the types and exports we're defining in our `package.json` files are usable under a variety of different Node resolution systems.


kanadgupta (approved):
this `--profile` flag is a g*dsend

PR #1029, opened 2025-05-16
feat: debugging mode

### LINKED ISSUE #789: Debugging mode
### Discussed in https://github.com/readmeio/api/discussions/545

<div type='discussions-op-text'>

<sup>Originally posted by **erunion** October 25, 2022</sup>
Because a lot of the work that `api` does behind the scenes can be a bit magical it'd be nice if there was some sort of debugging mode that you could enable that would log out the HAR that we use in the backend with [fetch-har](https://npm.im/fetch-har).

```js
const sdk = require('@api/petstore');

// leaf handling
sdk.debug();
sdk.operation().then(...)
```

```js
const sdk = require('@api/petstore');

sdk.debug().operation().then(...)
```

Perhaps the chained mode should only apply the debug mode to the operation it's attached to.</div>

AndriiAndreiev:
| 🚥 Resolves #789  |
| :------------------- |

## 🧰 Changes

This PR adds a debug() method to the API that enables a debugging mode to log HAR data, providing better visibility into what the API is doing under the hood. Since much of the API's behavior can feel a bit "magical," this helps make it more transparent and easier to troubleshoot.

Usage supports two styles:

```
sdk.debug();
sdk.operation().then(...);

sdk.debug().operation().then(...);
```

The chained version `(sdk.debug().operation())` applies debug mode to just that single operation, while the top-level sdk.debug() enables it globally.

## 🧬 QA & Testing

Import the generated SDK and enable the debug mode globally or per call using the chained version.

- Without debug(): No HAR logs should be printed.
- With global debug(): All subsequent SDK calls should print HAR logs.
- With chained debug(): Only that specific call should log HAR data.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5971,6 +5971,7 @@
       "version": "5.4.1",
       "resolved": "https://registry.npmjs.org/chalk/-/chalk-5.4.1.tgz",
       "integrity": "sha512-zgVZuo2WcZgfUEmsn6eO3kINexW8RAE4maiQ8QNs8CtpPCSyMiYsULR3HQYkm3w8FIA3SberyMJMSldGsW+U3w==",
+      "license": "MIT",
       "engines": {
         "node": "^12.17.0 || ^14.13 || >=16.0.0"
       },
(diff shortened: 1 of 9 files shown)


erunion (commented):
This is awesome. How much work would it be to add some tests for?

PR #1030, opened 2025-05-19
refactor: shared package

AndriiAndreiev:
## 🧰 Changes

- Created a new shared package (@readme/api-shared) inside the monorepo to host common utilities.
- Extracted the logger implementation from `api` into this shared package to enable reuse across multiple packages.

This setup is especially needed to support logging in **debugging mode** across the monorepo, including adding logging capabilities to the `api-core` package. (PR #1029). The shared package will also serve as a place to add new features/utilities intended for use in multiple packages throughout the monorepo.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -11,6 +11,7 @@
     {
       "files": ["packages/**/test/**/*"],
       "rules": {
+        "readme/no-wildcard-imports": "off",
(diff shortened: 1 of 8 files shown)


erunion:
@AndriiAndreiev why not move these shared utilities into `core` and then have `httpsnippet-client-api` depend on that?

PR #1061, opened 2025-08-16
feat: enhance ConfigOptions to include init parameter for fetch requests

### LINKED ISSUE #1046: Allow passing options to fetch
Looks like this should be simple to implement and absolutely required for my use case, as I need to be able to pass an `agent` property to the fetch options but it seems like this is not currently allowed. Looking at the code all I need is to be able to add custom values to the init (type RequestInit) property (possibly through a core.config entry) and it should pass through.

https://github.com/readmeio/api/blob/685212b1f6c4b573b0b4affd0c6f0c2b243f0bea/packages/core/src/index.ts#L100-L110

Magi1053:
| 🚥 Resolves #1046 |
| :------------------- |

## 🧰 Changes

This PR adds support for passing custom init options to the fetch request in the API core.
* The ConfigOptions interface now includes an optional init property, allowing users to specify additional RequestInit options.
* The APICore class now merges config.init into the fetch request's init object, ensuring user-supplied options are respected (except for signal, which is managed internally).

## 🧬 QA & Testing

1. Set up an instance of APICore and pass custom init options (e.g., custom headers or credentials) via setConfig({ init: { ... } }).
2. Make a request using fetch or fetchOperation.
3. Verify that the custom init options are present in the outgoing request (except for signal).
4. Ensure that the request still respects the internal timeout and abort logic.
5. Run the existing test suite to confirm no regressions.

### DIFF
--- a/packages/api/src/core/index.ts
+++ b/packages/api/src/core/index.ts
@@ -22,6 +22,11 @@ export interface ConfigOptions {
    * in milliseconds.
    */
   timeout?: number;
+  /**
+   * Additional arguments to pass to the `init` object used in fetch requests.
+   * These will be shallow-merged with internal options. The `signal` property is always managed internally and cannot be overridden.
+   */
+  init?: Omit<RequestInit, 'signal'>;
 }
 
 export interface FetchResponse<status, data> {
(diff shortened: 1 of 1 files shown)


erunion:
closing in favor of https://github.com/readmeio/api/pull/1183, thank you!

